PR #5, opened 2021-03-05
Make sure both 'settings' and 'builders' are in the YAML.

clalancette:
That is, don't fall back to empty defaults for them.  This
is important in case users provide their own YAML configuration,
but have a typo in them (like 'biulders' instead of 'builders').
With the default of an empty dict, the builders would just be
skipped completely, and could lead to hard-to-debug issues.
At least with this PR it would be louder.

@wjwwood @tfoote This seems fairly harmless to me, but an opinion on whether this is the right way to go is appreciated.

### DIFF
--- a/rosdoc2/verbs/build/parse_rosdoc2_yaml.py
+++ b/rosdoc2/verbs/build/parse_rosdoc2_yaml.py
@@ -61,8 +61,23 @@ def parse_rosdoc2_yaml(yaml_string, build_context):
             f"Error parsing file '{file_name}', in the second section, "
             f"expected something like dict{{settings: <tool settings>, builders: <builders>}}, "
             f"got a '{type(config)}' instead")
-    settings_dict = config.get('settings', {})
-    builders_dict = config.get('builders', {})
+
+    if 'settings' not in config:
+        raise ValueError(
+            f"Error parsing file '{file_name}', in the second section, "
+            f"expected a 'settings' key")
+    settings_dict = config['settings']
+    if not isinstance(settings_dict, dict):
+        raise ValueError(
+            f"Error parsing file '{file_name}', in the second section, value 'settings', "
+            f"expected a dict{{output_dir: build_settings, ...}}, "
+            f"got a '{type(builders_dict)}' instead")
+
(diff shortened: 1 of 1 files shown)


wjwwood (approved):
I guess this is ok, I think originally I was trying to let users specify as little as possible and let it be valid, but maybe this is just better.

An alternative would be to check for keys that are not `settings` or `builders`, and error on that. That would catch your `biulders` issue but still allow users to skip sections that they did not want to fill out. But I really don't mind.

tfoote:
The biggest question is at some point do we see the default builder(s) changing at some point down the line. If so having everyone hardcoded with the current defaults will make that transition much harder. On the flip side, they will keep working if the new builders need a migration. I agree that we should have an error the extra key rejection could cover that case, in which case I'd probably lean towards suggesting that approach instead of adding more required things that might need migrating in the future. We should probably also make sure that the "default builders (x y z) selected due to no config element in the yaml" message be printed promenantly for visibility/discoverability.

clalancette:
Just to be clear, this PR doesn't change anything about the default behavior, just the behavior when the user specifies a config file.  Some examples:

| Scenario | Before this PR | after this PR |
| ------------ | ------------------- | ---------------- |
| No rosdoc2.yaml specified | Use default builders (doxygen and sphinx) | Use default builders (doxygen and sphinx) |
| rosdoc2.yaml specified with 'settings' and 'builders' | Use builders specified by user | Use builders specified by user |
| rosdoc2.yaml specified with no 'builders' and no 'settings' | Use default settings, don't run any builders | Error |
| rosdoc2.yaml specified with 'builders' but no 'settings' | Use builders specified by user, and default for settings | Error |
| rosdoc2.yaml specified with 'settings' but no 'builders' | Don't run any builders | Error |

Because of that, I think it is still easy to change the default set of builders in the future.  The only thing this PR makes is so that once you specify something in `rosdoc2.yaml`, you have to specify everything.  We could think about making this a little less explicit, but I feel like that is easy enough to do later.  So I'm going to go ahead and merge this for now, thanks for the conversation.

PR #7, opened 2021-03-08
Remove unused setup_doc_build_prefix.py

clalancette:

@wjwwood This looks like an attempt to use code to generate the config files, but it also looks like it is currently unused.  Is it OK to remove?


wjwwood (approved):
Right it is vestigial, I was keeping it around to canabalize from, but I think we can safely remove it now.

PR #8, opened 2021-03-08
Change the YAML builder configuration arguments.

clalancette:
Make the builders a list, where the key for each one is the
type of builder and the values is a dictionary of required
and optional keys.

This may be the most important one that we agree on, since it fundamentally changes the configuration file that users provide (and thus it will be hard to change later).  Opinions welcome here.

### DIFF
--- a/rosdoc2/verbs/build/parse_rosdoc2_yaml.py
+++ b/rosdoc2/verbs/build/parse_rosdoc2_yaml.py
@@ -73,7 +73,7 @@ def parse_rosdoc2_yaml(yaml_string, build_context):
         if len(builder) != 1:
             raise ValueError(
                 f"Error parsing file '{file_name}', in the second section, each builder "
-                f"must have a 'type'")
+                f"must have exactly one key (which is the type of builder to use)")
         builder_name = next(iter(builder))
         builders.append(create_builder_by_name(builder_name, builder_dict=builder[builder_name], build_context=build_context))
 
--- a/rosdoc2/verbs/build/parse_rosdoc2_yaml.py
+++ b/rosdoc2/verbs/build/parse_rosdoc2_yaml.py
@@ -17,22 +17,6 @@ import yaml
 from .builders import create_builder_by_name
 
 
-def parse_builder_entry(output_directory, builder_dict, build_context):
-    """
-    Parse a single builder dictionary entry.
-    """
-    if 'builder' not in builder_dict:
-        keys = ', '.join(list(builder_dict.keys()))
-        raise ValueError(
-            f"Error parsing file '{build_context.configuration_file_path}', "
-            f"expected to find a 'builder' key but found only '[{keys}]'")
-    return create_builder_by_name(
-        builder_dict['builder'],
-        builder_dict=builder_dict,
-        output_dir=output_directory,
-        build_context=build_context)
-
-
 def parse_rosdoc2_yaml(yaml_string, build_context):
     """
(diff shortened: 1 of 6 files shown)


wjwwood (changes requested):
I don't think this is the right direction.

The problem with this is that it limits you to a single builder of each type, since the builder type is the key to the dictionary.

This is why the output directory is currently the key, to easily enforce that two builders do not share the same output directory.

However, I understand why it's feels odd. Also, after working on it more, I'm starting to challenge my own idea that "builders should never have the same output directory".

So if we change it I would suggest one of these two options instead.

Either we put the "name" as the key, e.g.:

```yaml
builders:
  '{package_name} Public C/C++ API':
    builder: doxygen
    output_dir: 'generated_doxygen'
  '{package_name}':
    builder: sphinx
    output_dir: ''
    # ...
```

The problem with this is that since we have some place holders like `{package_name}` in there, they could end up being the same later... but that's no better or worse than now.

Or we could make them a list of dictionaries, each organized like you proposed here:

```yaml
builders: [
  - sphinx: {
    name: '{package_name}'
    output_dir: ''
  }
  - doxygen: {
    name: '{package_name} Public C/C++ API'
    output_dir: 'generated/doxygen'
  }
]
```

Anyway, that's my thinking.

clalancette:
Excellent points, thanks William.

I like your idea about the list of dictionaries, though I would change it slightly:

```
builders: [
  - {
    type: 'doxygen',
    name: '{package_name} Public C/C++ API'
    output_dir: 'generated/doxygen'
  }
  - {
    type: 'sphinx',
    name: '{package_name}',
    output_dir: ''
  }
]
```

I'm going to go ahead and code that up.

wjwwood:
Ok, sounds good. 

Also I should say that the reason the key being a builder was concerning for me was that I have been toying with the idea that the landing page should just be another Sphinx builder invocation. But I'm not firm on that yet.

clalancette:
> I like your idea about the list of dictionaries, though I would change it slightly:

Actually, after looking at it, I decided that your idea was best, so I went with that.

> Also I should say that the reason the key being a builder was concerning for me was that I have been toying with the idea that the landing page should just be another Sphinx builder invocation. But I'm not firm on that yet.

Your idea makes perfect sense, users may want to run things more than once for some reason.  I've now tested that you can specify multiple of the same 'type' of builder with the new code now, and it works.  Please take another look when you get a chance.

rosdoc2/verbs/build/parse_rosdoc2_yaml.py
```diff
@@ -77,14 +61,20 @@ def parse_rosdoc2_yaml(yaml_string, build_context):
...
+            f"got a '{type(builders_list)}' instead")
 
     builders = []
-    for output_directory, entry in builders_dict.items():
-        builders.append(parse_builder_entry(output_directory, entry, build_context))
+    for builder in builders_list:
+        if len(builder) != 1:
+            raise ValueError(
+                f"Error parsing file '{file_name}', in the second section, each builder "
+                f"must have a 'type'")
```
wjwwood, line 76: Is this the right error message?

wjwwood (approved):
One question, but otherwise lgtm

rosdoc2/verbs/build/parse_rosdoc2_yaml.py
clalancette: Yeah, that was wrong.  I've cleaned it up in 5b8f2a8.

clalancette:
Thanks for the review, going ahead and merging this one.

PR #10, opened 2021-03-09
Remove --{build,install}-directory arguments.

clalancette:
As far as I can tell, they are unused at the moment.  Removing
them simplifies running the tool for end-users.

Any opinions on whether we need this or not is welcome.


clalancette:
This was superseded by #13, so closing.

PR #11, opened 2021-03-09
enable automatic intersphinx mapping extension between packages

wjwwood:
This is in the same fashion as the doxygen tag files, using a `.location.json` file along side the `objects.inv` to handle sphinx when it is in a subdirectory of the documentation output directory (not the default).


wjwwood:
So, it seems to work between `rcutils` and `rcl`:

https://giphy.com/gifs/s0UfnByYjOiOLdiToK/html5

But that's the extent of my testing. I think it should work though, since it's based on the Doxygen one and I tested it more.

wjwwood:
Relative links are not going to work, because depending on the thing being documenting, the number of times you go up will differ. I used something like `file:///abs/path/to/docs_output` as my `--base-url`.

PR #12, opened 2021-03-10
Switch to Read-the-docs theme.

clalancette:
This makes it consistent with the documentation hosted at
https://docs.ros.org/en/foxy

As we discussed earlier, switch to RTD theme.

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -88,8 +88,9 @@ if rosdoc2_settings.get('enable_exhale', True):
     }})
 
 if rosdoc2_settings.get('override_theme', True):
+    extensions.append('sphinx_rtd_theme')
+    html_theme = 'sphinx_rtd_theme'
     print(f"[rosdoc2] overriding theme to be '{{html_theme}}'", file=sys.stderr)
-    html_theme = 'classic'
 
 if rosdoc2_settings.get('automatically_extend_intersphinx_mapping', True):
     print(f"[rosdoc2] extending intersphinx mapping", file=sys.stderr)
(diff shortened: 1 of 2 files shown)


rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -88,8 +88,9 @@ def ensure_global(name, default):
     }})
 
 if rosdoc2_settings.get('override_theme', True):
+    extensions.append('sphinx_rtd_theme')
+    html_theme = 'sphinx_rtd_theme'
     print(f"[rosdoc2] overriding theme to be '{{html_theme}}'", file=sys.stderr)
```
wjwwood, line 93: nitpick: the other stanza's have the print before the activity

```suggestion
    print(f"[rosdoc2] overriding theme to be '{{html_theme}}'", file=sys.stderr)
    extensions.append('sphinx_rtd_theme')
    html_theme = 'sphinx_rtd_theme'
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -171,7 +172,7 @@ def ensure_global(name, default):
 #
 ## rosdoc2 will override the theme, but you may set one here for running Sphinx
 ## without the rosdoc2 tool.
-html_theme = 'alabaster'
+html_theme = 'sphinx_rtd_theme'
```
wjwwood, line 175: So if the user copies this and then runs without rosdoc2, this will fail I think, since it isn't in the extensions?

wjwwood: Or does it just need to be installed? in which case adding it to the extensions above isn't right?

rosdoc2/verbs/build/builders/sphinx_builder.py
clalancette: I don't think this works in this case, though, since we need to set the `html_theme` variable before using it in the print.

rosdoc2/verbs/build/builders/sphinx_builder.py
clalancette: No, we need to add it to the extensions.  Good catch.

rosdoc2/verbs/build/builders/sphinx_builder.py
wjwwood: Oh yeah 🙃

wjwwood (approved):

PR #20, opened 2021-07-23
Fix toctree warnings and render toctree in the HTML sidebar

aprotyas:
This PR fixes a warning about the library API document not being included in the document, and also fixes the sidebar rendering (table of content shows up in the sidebar correctly).

These two issues are connected because the Sphinx html theme uses the `:toctree:` directive to create the sidebar.

References:
* Exhale, link to generated documentation API: https://exhale.readthedocs.io/en/latest/usage.html#make-your-documentation-link-to-the-generated-api
* Sphinx rtd theme, table of content options: https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#table-of-contents-options

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -228,7 +228,6 @@ C++ API
 .. toctree::
    :maxdepth: 2
 
-   README <api/program_listing_file_README.md.rst>
    api/library_root
    Full API <api/unabridged_api>
    File structure <api/unabridged_orphan>
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -170,6 +170,15 @@ master_doc = 'index'
 ## without the rosdoc2 tool.
 html_theme = 'sphinx_rtd_theme'
 
+html_theme_options = {{
+    # Toc options
+    'collapse_navigation': False,
+    'sticky_navigation': True,
+    'navigation_depth': -1,
+    'includehidden': True,
+    'titles_only': False,
+}}
+
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
@@ -216,9 +225,13 @@ index_rst_template = """\
 C++ API
 =======
 
-:doc:`api/library_root`
+.. toctree::
+   :maxdepth: 2
 
-.. doxygenindex::
+   README <api/program_listing_file_README.md.rst>
+   api/library_root
+   Full API <api/unabridged_api>
+   File structure <api/unabridged_orphan>
 
 Indices and Search
 ==================


rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -216,9 +225,13 @@ def ensure_global(name, default):
 C++ API
 =======
 
-:doc:`api/library_root`
+.. toctree::
+   :maxdepth: 2
 
-.. doxygenindex::
+   README <api/program_listing_file_README.md.rst>
```
aprotyas, line 231: For reviewers, this change assumes that a README file is present in the project root. Is this a reasonable assumption?

clalancette: I would say we shouldn't assume it.  While it is really common, it is not universal (nor is it universally README.md).  Since the user can provide their own Doxyfile to include whatever they want, there's at least an option to have the README included.

That's the long-winded way of saying that I Think we should remove this :).

aprotyas: That seems reasonable. I'll remove the README tag from the `toctree`.

This can open discussion on the package description in the `index.rst` file.

Should the package description strictly be pulled from the package's `package.xml` manifest file? I'm imagining a situation where the author's have a detailed project description in some form of a README file (not necessarily README.md). In that case, `project.description` could be replaced with a separate `project_description` template variable. This variable would be populated with:
* If the project contains a Doxyfile, and the Doxyfile provides an `USE_MDFILE_AS_MAINPAGE` argument, then `project_description` would be an include directive with that MDFILE.
* The contents of the package description from `package.xml`
https://github.com/ros-infrastructure/rosdoc2/blob/98cfe30c1945a68e1ce5cc009a43c13b14c1d58b/rosdoc2/verbs/build/builders/sphinx_builder.py#L210-L229

aprotyas: > That's the long-winded way of saying that I think we should remove this :).

Addressed in 5f82cbc.

wjwwood (approved):

rosdoc2/verbs/build/builders/sphinx_builder.py
wjwwood: > Should the package description strictly be pulled from the package's package.xml manifest file?

Maybe, if you open a pull request, or a hand written example of ideal output, with what you're thinking that would be better for discussion (something concrete to talk about). In general we need to improve the landing page a lot. But we wanted to get something up first. Ideally we'd have something like the "Package Summary" on the package pages on the ROS 1 wiki, e.g.:

https://wiki.ros.org/ros_comm

Is that what you're thinking about?

wjwwood: @aprotyas can we resolve this for the context of this pull request?

aprotyas: I've moved this conversation to #22 and can resolve this for the context of this pull request.

aprotyas:
@wjwwood can you merge the PR?

PR #21, opened 2021-07-23
Add markdown parsing support to `rosdoc2`

aprotyas:
This PR introduces markdown parsing support in two separate ways.

1. If there are markdown files in the `Sphinx` project where the
   `SphinxBuilder` object is run, previously those files were treated as
   restructured text files. By adding the `myst-parser` to the list of
   extensions and by registering the `.md` source suffix to correspond
   to `markdown` files, the appropriate parser will be called for any
   markdown files in the `Sphinx` project tree.
   This is the less likely situation because when a default `Sphinx`
   project tree is generated, no files from the existing package
   repository is transferred.
2. If a project's Doxygen configuration includes any markdown files,
   then `exhale` provides those files with a `file_<>.rst` and a
   `program_listing_<>.rst` page, where the latter is included as a
   codeblock in the former. This is not ideal - as exhale should be
   providing such files with `:doxygenpage:` directives [(issue raised
   here)](https://github.com/svenevs/exhale/issues/111).
   Regardless, the appropriate Pygments lexer (`md`) needs to be
   registered for markdown code blocks, which this patch addresses.

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -74,6 +74,8 @@ if rosdoc2_settings.get('enable_exhale', True):
     print('[rosdoc2] enabling exhale', file=sys.stderr)
     extensions.append('exhale')
     ensure_global('exhale_args', {{}})
+    
+    from exhale import utils
     exhale_args.update({{
         # These arguments are required.
         "containmentFolder": "{user_sourcedir}/api",
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -87,9 +87,9 @@ if rosdoc2_settings.get('enable_exhale', True):
         "exhaleExecutesDoxygen": False,
         # Maps markdown files to the "md" lexer, and not the "markdown" lexer
         # Pygments registers "md" as a valid markdown lexer, and not "markdown"
-        "lexerMapping": {{r".*\.md": "md",}},
+        "lexerMapping": {{r".*\.(md|markdown)$": "md",}},
         # This mapping will work when `exhale` supports `:doxygenpage:` directives
-        # Check `svenevs/exhale#111`
+        # Check https://github.com/svenevs/exhale/issues/111
         "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
         lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
     }})
@@ -178,6 +178,7 @@ master_doc = 'index'
 source_suffix = {{
     '.rst': 'restructuredtext',
     '.md': 'markdown',
+    '.markdown': 'markdown',
 }}
 
 # -- Options for HTML output -------------------------------------------------
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -85,6 +85,13 @@ if rosdoc2_settings.get('enable_exhale', True):
         # TIP: if using the sphinx-bootstrap-theme, you need
         # "treeViewIsBootstrap": True,
         "exhaleExecutesDoxygen": False,
+        # Maps markdown files to the "md" lexer, and not the "markdown" lexer
+        # Pygments registers "md" as a valid markdown lexer, and not "markdown"
+        "lexerMapping": {{r".*\.md": "md",}},
+        # This mapping will work when `exhale` supports `:doxygenpage:` directives
+        # Check `svenevs/exhale#111`
+        "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
+        lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
     }})
 
 if rosdoc2_settings.get('override_theme', True):
@@ -101,6 +108,14 @@ if rosdoc2_settings.get('automatically_extend_intersphinx_mapping', True):
     ensure_global('intersphinx_mapping', {{
         {intersphinx_mapping_extensions}
     }})
+
+if rosdoc2_settings.get('support_markdown', True):
+    print(f"[rosdoc2] adding markdown parser", file=sys.stderr)
+    # The `myst_parser` is used specifically if there are markdown files
+    # in the sphinx project
+    # TODO(aprotyas): Migrate files under the `include` tag in the project's Doxygen
+    # configuration into the Sphinx project tree used to run the Sphinx builder in.
+    extensions.append('myst_parser')
 """
 
 default_conf_py_template = """\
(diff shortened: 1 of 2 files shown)


rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -85,6 +85,13 @@ def ensure_global(name, default):
         # TIP: if using the sphinx-bootstrap-theme, you need
         # "treeViewIsBootstrap": True,
         "exhaleExecutesDoxygen": False,
+        # Maps markdown files to the "md" lexer, and not the "markdown" lexer
+        # Pygments registers "md" as a valid markdown lexer, and not "markdown"
+        "lexerMapping": {{r".*\.md": "md",}},
+        # This mapping will work when `exhale` supports `:doxygenpage:` directives
+        # Check `svenevs/exhale#111`
+        "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
+        lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
```
aprotyas, line 94: For reviewers, the [`customSpecificationsMapping`]() argument allows users to:

> For a given input ``kind``, return the list of reStructuredText specifications for the associated Breathe directive.

Essentially, this mapping should allow `:doxygenpage:` inputs to only show the markdown text (`:content-only:`), rather than redirecting the user to a separate html page for the markdown text.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -85,6 +85,13 @@ def ensure_global(name, default):
         # TIP: if using the sphinx-bootstrap-theme, you need
         # "treeViewIsBootstrap": True,
         "exhaleExecutesDoxygen": False,
+        # Maps markdown files to the "md" lexer, and not the "markdown" lexer
+        # Pygments registers "md" as a valid markdown lexer, and not "markdown"
+        "lexerMapping": {{r".*\.md": "md",}},
+        # This mapping will work when `exhale` supports `:doxygenpage:` directives
+        # Check `svenevs/exhale#111`
```
wjwwood, line 92: I suggest using the full GitHub url here, or mention it is on GitHub, otherwise you don't know where it is hosted, could be GitLab for instance.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -85,6 +85,13 @@ def ensure_global(name, default):
         # TIP: if using the sphinx-bootstrap-theme, you need
         # "treeViewIsBootstrap": True,
         "exhaleExecutesDoxygen": False,
+        # Maps markdown files to the "md" lexer, and not the "markdown" lexer
+        # Pygments registers "md" as a valid markdown lexer, and not "markdown"
+        "lexerMapping": {{r".*\.md": "md",}},
```
wjwwood, line 90: Consider adding `.markdown` too.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -160,6 +175,10 @@ def ensure_global(name, default):
 
 master_doc = 'index'
 
+source_suffix = {{
+    '.rst': 'restructuredtext',
+    '.md': 'markdown',
```
wjwwood, line 180: Here too, `.markdown`.

wjwwood (approved):
A few small fixups suggested, but otherwise lgtm.

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Addressed in 2317689.

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Addressed in 2317689.

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Full link added in 2317689.

aprotyas:
@wjwwood can you merge this PR?

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -85,6 +87,13 @@ def ensure_global(name, default):
         # TIP: if using the sphinx-bootstrap-theme, you need
         # "treeViewIsBootstrap": True,
         "exhaleExecutesDoxygen": False,
+        # Maps markdown files to the "md" lexer, and not the "markdown" lexer
+        # Pygments registers "md" as a valid markdown lexer, and not "markdown"
+        "lexerMapping": {{r".*\.(md|markdown)$": "md",}},
+        # This mapping will work when `exhale` supports `:doxygenpage:` directives
+        # Check https://github.com/svenevs/exhale/issues/111
+        "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
+        lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
```
wjwwood, line 96: Can you indent here, it looks strange to have these two lines be indented to the same level.

wjwwood (commented):
Small comment, otherwise lgtm

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Indented in 4768a1b.

FYI, I've also commented out the actual specification remapping because as it stands, this remapping causes unexpected behavior for any other directive (such as `:doxygenclass:`) not explicitly listed in the remapping. This can be addressed once support for `:doxygenpage:` directives is formally introduced into `exhale`.

If needed, I can just remove the commented lines out instead. In any case, those statements shouldn't execute now.

wjwwood: Ok

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -85,6 +87,14 @@ def ensure_global(name, default):
...
         # "treeViewIsBootstrap": True,
         "exhaleExecutesDoxygen": False,
+        # Maps markdown files to the "md" lexer, and not the "markdown" lexer
+        # Pygments registers "md" as a valid markdown lexer, and not "markdown"
+        "lexerMapping": {{r".*\.(md|markdown)$": "md",}},
+        # This mapping will work when `exhale` supports `:doxygenpage:` directives
+        # Check https://github.com/svenevs/exhale/issues/111
+        # TODO(aprotyas): Uncomment the mapping below once the above issue is resolved.
+        #"customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
+        #    lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
```
wjwwood, line 97:
```suggestion
        # "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
        #     lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
```

Issue #19, opened 2021-07-20
packages without doxygen don't generate content


tfoote:
building geometry2 only results in uploads of tf2_ros

https://build.ros2.org/view/Rdoc/job/Rdoc__geometry2__ubuntu_focal_amd64/4/console

This may be the only package with doxygen setup so it may be just that the sphinx fallback isn't working. 
Other packages in geometry_tutorials and common interfaces don't generate anything: 

https://build.ros2.org/view/Rdoc/job/Rdoc__geometry_tutorials__ubuntu_focal_amd64/4/console
https://build.ros2.org/view/Rdoc/job/Rdoc__common_interfaces__ubuntu_focal_amd64/3/console

And filters passes: https://build.ros2.org/view/Rdoc/job/Rdoc__filters__ubuntu_focal_amd64/4/console

PR #24, opened 2021-07-29
Add `rosdoc2` option to specify `primary_domain`

aprotyas:
This option should enable `sphinx` to land on reasonable defaults for
code highlighting Pygments lexers, as well as other markup behaviors,
depending on the domain of a package (defaults to `cpp`).


clalancette (commented):
So this PR would allow a user to provide a rosdoc2.yaml file to override and get Python highlighting.  If they did *not* do this, then they would get C++ highlighting, even on Python packages.  Is that a correct assessment of the change?

aprotyas:
> So this PR would allow a user to provide a rosdoc2.yaml file to override and get Python highlighting. If they did _not_ do this, then they would get C++ highlighting, even on Python packages. Is that a correct assessment of the change?

That's right. There's an assumption here about the frequency of C++ packages over Python packages, which is not correct in hindsight.
If the `primary_domain`/`highlight_domain` options are not set, the default value is `default`, which is a pseudo-Python3 syntax highlighting. I don't know if that's ideal either as a default case for `rosdoc2`.

aprotyas:
> So this PR would allow a user to provide a rosdoc2.yaml file to override and get Python highlighting. If they did _not_ do this, then they would get C++ highlighting, even on Python packages. Is that a correct assessment of the change?

That's right. There's an assumption here about the frequency of C++ packages over Python packages, which is not correct in hindsight.
If the `primary_domain`/`highlight_domain` options are not set, the default values are [`py`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-primary_domain) and [`default`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-highlight_language) (basically Python3) respectively. I don't know if that's ideal either as a default case for `rosdoc2`. If they are set to `None`, then of course there is no highlighting.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -119,6 +119,16 @@ def ensure_global(name, default):
     # TODO(aprotyas): Migrate files under the `include` tag in the project's Doxygen
     # configuration into the Sphinx project tree used to run the Sphinx builder in.
     extensions.append('myst_parser')
+
+# if no `primary_domain` option provided, defaults to 'cpp'
+if rosdoc2_settings.get('primary_domain', 'cpp'):
+    print(f"[rosdoc2] setting primary domain to '{{rosdoc2_settings.get('primary_domain')}}'",
+        file=sys.stderr)
+    # Tell sphinx what the primary language being documented is.
+    primary_domain = rosdoc2_settings.get('primary_domain')
```
wjwwood, line 128: What if the user already set `primary_domain` or `highlight_language`? We'd be overwriting them. I think it would be better to only set these if `primary_domain` is explicitly in `rosdoc2_settings`, or if the user didn't specify `primary_domain` or `highlight_language`.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -119,6 +119,16 @@ def ensure_global(name, default):
     # TODO(aprotyas): Migrate files under the `include` tag in the project's Doxygen
     # configuration into the Sphinx project tree used to run the Sphinx builder in.
     extensions.append('myst_parser')
+
+# if no `primary_domain` option provided, defaults to 'cpp'
+if rosdoc2_settings.get('primary_domain', 'cpp'):
```
wjwwood, line 124: This condition is a bit weird? Basically it would only be false if the user explicitly did `rosdoc2_settings = {'primary_domain': None}` or something like that.

aprotyas:
In the case that the user specifies neither a `primary_domain`/`highlight_language` in their `conf.py` configuration nor a `primary_domain` in `rosdoc2_settings` (i.e. it is not the case that the user's setting will be overwritten), is there a reasonable default for the domain/language that `rosdoc2` could/should configure? The `sphinx` default value for these options is `None`/`py` respectively, which is essentially a pseudo-Python3 syntax. I'm leaning towards a `cpp` default, but I can't back that up with the relative frequency of `ament_cmake` packages over that of `ament_python` packages.

Thoughts?

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: I concur, I've defaulted to `None` rather than `cpp` for this condition in ac4a5ea595ffd00a64d99c7a7eed4dc62bc773a9.

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: With the changes in ac4a5ea595ffd00a64d99c7a7eed4dc62bc773a9, the last talking point is about a reasonable default for the domain/highlighting - check https://github.com/ros-infrastructure/rosdoc2/pull/24#issuecomment-893979131. For example, if `cpp` were the default provided by `rosdoc2` in case the user didn't specify `primary_domain` or `highlight_language` in their `conf.py` configuration, the following snippet should do the trick:

```python
ensure_global('primary_domain', 'cpp')
ensure_global('highlight_language', 'cpp')

if rosdoc2_settings.get('primary_domain'):
    ...
```

Thoughts?

wjwwood: Hmm, I wouldn't default to `cpp` I think. The user can always explicitly label the language of code in their doc blocks right? Or does this cover a case where that doesn't work?

If we do pick a default other than the sphinx default, I'd say we should do `cpp` if it is an `ament_cmake` package and `py` or do nothing if it is an `ament_python` package.

aprotyas: > Or does this cover a case where that doesn't work?

Off the top of my head, I don't think this would cover any case where the user couldn't explicitly label the code block.

> I'd say we should do `cpp` if it is an `ament_cmake` package and `py` or do nothing if it is an `ament_python` package.

Alright, I will keep this in mind as I'm working on #19.

aprotyas:
With ac4a5ea, the current behavior is to not change the primary domain setting in any way unless explicitly specified. However, having extracted build type information in #28, it would make more sense to set the primary domain to `cpp`/`py` depending on whether the build type is `ament_cmake`/`ament_python` (suggestion in https://github.com/ros-infrastructure/rosdoc2/pull/24#discussion_r683977435). This information can be format mapped to `rosdoc2_wrapping_conf_py_template` through a template variable.

Thoughts?

wjwwood:
> However, having extracted build type information in #28, it would make more sense to set the primary domain to cpp/py depending on whether the build type is ament_cmake/ament_python (suggestion in #24 (comment)).

Yes I think so, I was imaging `cpp` for `ament_cmake` and not setting anything (using the default) for `ament_python`. But if the standard practice for Python projects is to manually set it to `py` then that would be ok for `ament_python` as well. Basically I think `ament_python` should do whatever is "normal" for a plain python package.

Issue #22, opened 2021-07-29
Improve package landing page


aprotyas:
## What?

This issue will target discussions about general improvements to a package's API documentation's landing page, housing the conversation in https://github.com/ros-infrastructure/rosdoc2/pull/20#discussion_r675916581 somewhere outside that PR. Currently, the landing page simply draws the package description from the package manifest (`package.xml`), which is usually just a one-liner about the package; this description is followed by the API listing.

Ideally, the landing page for a package's API documentation should provide a longer description - and hence more context/information about the package, something usually found in an auxiliary documentation file, such as a README.

<details>
<summary>Screenshot of <code>rcpputils</code> documentation for visual reference</summary>

![Screenshot 2021-07-28 at 19-30-30 rcpputils — rcpputils 2 2 0 documentation](https://user-images.githubusercontent.com/34530011/127429620-e14d495b-0d0f-47ad-b83c-d6b773ba9991.png)

</details>

## How?

My idea of improving the index page was to somehow inject/render the content of an **optional**, **user-configurable** documentation file in place of the `package.description` mapped into the `index.rst` template by the `SphinxBuilder`.

### Implementation

There are a few methods to consider. To use any of these methods, the user would have to add a `{use_mdfile_as_mainpage: <document>}`  option for `SphinxBuilder` in the package's `rosdoc2.yaml` file.

1. Using markdown parsers with `include` directives: Once #21 is merged, the package description can be changed to an include directive with the specified file. This file would then just get parsed into `rst` format by the `myst-parser`. **Problem**: Relative file links will not work, because they link to exactly that file name in the host. On the other hand, the documents rendered by `sphinx` now don't even generate relative file links, let alone be broken - because documentation files are just rendered as program listings, similar to `.cpp/.hpp` files.

<details>
<summary>Proposed changes in <code>sphinx_builder.py</code></summary>

```python3
    def generate_default_project_into_directory(self, directory):                                         
        if use_mdfile_as_mainpage:                                                                           
            package_description = f".. include:: {source-directory}/<documentation-file>"
        else:                                                                                       
            package_description = package.description                                               
        ...                                               
        template_variables = {                                                                      
            ...                                    
            'package_description': package_description,    
```

</details>

2. 
3. Test

clalancette:
First of all, thanks for starting this discussion.  I agree that it would be really nice to improve the landing page.

However, I think we can do some more things automatically here.  Looking at your original mockup, it looks like we have a structure that looks something like:

[...]

PR #25, opened 2021-07-29
Reorganized `SphinxBuilder` for markdown parsing

aprotyas:
* Removed a stray TODO comment from an experimental idea for #22.
* Moved the `lexerMapping` changes to `exhale` arguments into the
  conditional `support_markdown` block, so that the appropriate Pygments
  lexer is only registered if support for markdown was enabled.
* Added longer description of the `support_markdown` option.


rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -116,9 +108,19 @@ def ensure_global(name, default):
     print(f"[rosdoc2] adding markdown parser", file=sys.stderr)
     # The `myst_parser` is used specifically if there are markdown files
     # in the sphinx project
-    # TODO(aprotyas): Migrate files under the `include` tag in the project's Doxygen
-    # configuration into the Sphinx project tree used to run the Sphinx builder in.
     extensions.append('myst_parser')
+    # If markdown support is enabled, the appropriate Pygments lexer must
+    # be registered for `.md`/`.markdown` file suffixes
+    exhale_args.update({{
```
clalancette, line 114: Will this work if `exhale_args` is `False` but `support_markdown` is `True`?

aprotyas: Good catch. This will not work if `enable_exhale` is `False`. This could be a way to check for that but maintain the logical structure.
```suggestion
    if rosdoc2_settings.get('enable_exhale'):
        exhale_args.update({{
```

aprotyas: Addressed in 6281b49.

An alternative to that patch would be to invoke `ensure_global` on `exhale_args`, but that may create a needless copy of `exhale_args` incase `enable_exhale` is false.

aprotyas:
Pinging @clalancette for a re-review following 6281b49, where `exhale_args` is now only updated if `enable_exhale` was set to true.

clalancette (approved):

PR #28, opened 2021-08-06
Add support for python-only projects

aprotyas:
Previously, `rosdoc2` would not generate any documentation for projects without doxygen - more concretely, projects with the `ament_python` build type - as reported in https://github.com/ros-infrastructure/rosdoc2/issues/19.

This PR attempts to resolve that issue. In offline discussions with @clalancette and @wjwwood, we decided that `rosdoc2` can exhibit different behavior for different build types. See [this comment](https://github.com/ros-infrastructure/rosdoc2/compare/main...aprotyas:aprotyas/build-python-pkg?expand=1#diff-c4e6990ce42081aecb92b47510d0be85b6857efafd406971d863f793379341c0R54-R62) for an outline of said behaviors.

Specifically, this PR:
- Adds a mandatory `build_type` configuration option, which defaults to the `build_type` obtained from the package.xml file. The set of available `build_type`s is [`ament_cmake`, `cmake`, `ament_python`].
- Change behavior of the `doxygen_builder` and `breathe`/`exhale` extensions for `ament_python` build types: For this build type, `doxygen` does not need to be invoked whatsoever. Similarly, the `breathe`/`exhale` extensions to `sphinx` need not be registered since there is no `doxygen`-generated tag file to process.
- Create `toctree` entries for the default `index.rst` specific to each build type.
- Add an optional `python_source` configuration option. This option is represented as a path relative to the package's `package.xml` file, and should really only be used for packages that do not follow the standard Python package layout. This defaults to `[package_name]`.
- Execute the `sphinx-apidoc` tool for `ament_python` build types so that documentation for Python modules can be automatically generated before `sphinx-build` is invoked.


aprotyas:
Pinging @clalancette and @wjwwood for a review.

rosdoc2/verbs/build/inspect_package_for_settings.py
```diff
@@ -44,6 +44,29 @@
...
+
+    ## Behaviors:
+    ## - 'ament_cmake': Runs 'doxygen_builder' and 'sphinx_builder'. The 'breathe' and
+    ##   'exhale' extension to 'sphinx' are also invoked.
+    ##   'ament_python': The 'doxygen_builder' is not invoked, and neither are the
+    ##   'breathe'/'exhale' extensions to 'sphinx'. Rather, the 'sphinx-apidoc' tool
+    ##   is invoked to generate '.rst' files for python modules.
+    ## - 'cmake': See 'ament_cmake'.
+    ## - 'mixed': Both the 'sphinx-apidoc' and the 'doxygen'/'breathe'/'exhale' tools
+    ##   are run, this build type requires a separate 'python_source' setting.
```
aprotyas, line 62:
```suggestion
    ##   are run, this build type may require a separate 'python_source' setting.
```

rosdoc2/verbs/build/builders/doxygen_builder.py
```diff
@@ -149,6 +157,13 @@ def __init__(self, builder_name, builder_entry_dictionary, build_context):
             logger.info(f"Using user specified Doxyfile at '{self.doxyfile}'.")
 
     def build(self, *, doc_build_folder, output_staging_directory):
+        # If the build type is `ament_python`, there is no reason to run doxygen.
+        if self.build_context.build_type == 'ament_python':
```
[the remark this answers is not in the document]
wjwwood: Sorry, I hadn't understood the `mixed` option yet. So maybe this comment doesn't apply.

wjwwood (commented):
I'm not sure about this `mixed` setting. What I was imagining is a default behavior for `ament_cmake` (+ doxygen/exhale/breathe, - sphinx-apidoc), a different one for `ament_python` (- doxygen/exhale/breathe, + sphinx-apidoc), and nothing special for all other package build types, but any of them could manually enable/disable things.

So if an `ament_cmake` package wanted to build python API docs too, then they would just enable that explicitly in the `rosdoc2_settings` dict, rather than change their type to `mixed`.

I can see the advantage to more "configuration templates", but it just feels like overkill at the moment, since the 90% case will be either `ament_cmake` with only C/C++ or `ament_python` with only Python. Everyone else in the meantime would need some special configurations, but if we see a common pattern emerging, then we could add a new "doc build type". As it stands, do you know of any packages that would select `mixed` right now?

Another thing to consider is packages that generate messages, but since we're not building first at the moment, that's kind of depending on that. We may have to change the logic more when that happens because right now all packages that have messages need to be `ament_cmake` but also will generate Python code for those messages. But again, I think we should defer that concern for now.

clalancette:
> I'm not sure about this `mixed` setting. What I was imagining is a default behavior for `ament_cmake` (+ doxygen/exhale/breathe, - sphinx-apidoc), a different one for `ament_python` (- doxygen/exhale/breathe, + sphinx-apidoc), and nothing special for all other package build types, but any of them could manually enable/disable things.

Yeah, agreed on that.  The other problem with 'mixed' is that if we add another documentation type (say for `ament_java` packages), then it is no longer clear what 'mixed' refers to.  Let's just go for the `ament_cmake`/`ament_python` split for now, and we can incrementally add other types as needed.  And anyone who needs to do more than one thing in their package can manually configure it.

rosdoc2/verbs/build/builders/doxygen_builder.py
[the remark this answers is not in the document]
aprotyas: If we drop the `mixed` option and stick to the `ament_cmake`/`ament_python` split for now, would it be reasonable to change this condition to

```suggestion
        if 'cmake' not in self.build_context.build_type:  # ['cmake', 'ament_cmake']
```
?
I'm still not entirely sure about the `cmake` build type altogether, so maybe instead

```suggestion
        if self.build_context.build_type != 'ament_cmake'
```
if the pure `cmake` build type is not supported?

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -53,49 +65,50 @@ def ensure_global(name, default):
...
+            "exhaleExecutesDoxygen": False,
+            # Maps markdown files to the "md" lexer, and not the "markdown" lexer
+            # Pygments registers "md" as a valid markdown lexer, and not "markdown"
+            "lexerMapping": {{r".*\.(md|markdown)$": "md",}},
+            # This mapping will work when `exhale` supports `:doxygenpage:` directives
+            # Check https://github.com/svenevs/exhale/issues/111
+            # TODO(aprotyas): Uncomment the mapping below once the above issue is resolved.
+            # "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
+            #     lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
+        }})
```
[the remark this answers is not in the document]
aprotyas: I thought about doing it that way, but once these settings are evaluated (i.e. `breathe`/`exhale` are added to the list of `sphinx` extensions), there is no convenient way to turn off their invocation. We could manually remove these from the list of extensions, conditional on what the build type is, but that would be extra work no?

rosdoc2/verbs/build/builders/doxygen_builder.py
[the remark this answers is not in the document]
wjwwood: For the cmake build type and any other build type that we don't explicitly support here I'd say we just try to run Sphinx if there is a doc folder with a conf.py. Otherwise do nothing special. We can revisit that later, improve the story for cmake and other build types later.

aprotyas:
It seems like the `mixed` setting adds more ambiguity than it removes, particularly when considering the existence of other build types (such as `ament_java`). I can remove that from this PR, instead maintaining an `ament_cmake`/`ament_python` split only.

> As it stands, do you know of any packages that would select `mixed` right now?

The only such package I can think of off the top of my head is `message_filters`.

> So if an `ament_cmake` package wanted to build python API docs too, then they would just enable that explicitly in the `rosdoc2_settings` dict, rather than change their type to `mixed`.

By "enable that explicitly", do you mean there should be keys `run_sphinx_apidoc` and `run_doxygen` which we should account for when determining default behavior for a build type? For example, for an `ament_cmake` package with `run_sphinx_apidoc` set to true, `sphinx-apidoc` will be excecuted?

aprotyas:
Regardless of the `mixed` setting, I've seen some packages that do not house the package source under the `{package_name}` directory - for example, `tf2_ros_py`'s source is placed under `tf2_ros`. I added the optional `python_source` setting in the `rosdoc2_settings` dict for situations like that (check 6612754). Should this stay as is? Is there a better name (or a better place) for such an option?

rosdoc2/verbs/build/builders/doxygen_builder.py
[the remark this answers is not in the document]
aprotyas: In 195c286, I've bundled `cmake` build type's behavior with that of the `ament_cmake` build type, i.e. they behave identically. I believe that's a reasonable assumption for such build types.

aprotyas:
I've dropped the `mixed` build type setting in 2977f35. In fact, the `build_type` information is no longer configurable (d5b4da9), since I've added the optional settings `run_doxygen` and `run_sphinx_apidoc` (5feb103 and d467c73 respectively) which allow overriding of expected/default behaviors for a given build type. More explicitly, setting those options to true allows their namesake tool to be invoked regardless of build type - so, an `ament_cmake` package could generate Python API documentation if it set `run_sphinx_apidoc` to true.

I've also patched up the toctree entries to distinguish between Python/C++ API headings in e43f3ab, just a small QOL thing.

Thoughts?

rosdoc2/verbs/build/builders/doxygen_builder.py
```diff
@@ -77,6 +77,15 @@ def __init__(self, builder_name, builder_entry_dictionary, build_context):
 
         assert self.builder_type == 'doxygen'
 
+        # If the build type is not `ament_cmake/cmake`, there is no reason
+        # to create a doxygen builder.
+        if (self.build_context.build_type not in ['ament_cmake', 'cmake'] and
+                not self.build_context.run_doxygen):
+            logger.info(
+                f"The package build type is not 'ament_cmake/cmake', hence the"
+                f"'{self.builder_type}' builder is not generated")
```
clalancette, line 86:
```suggestion
        if (self.build_context.build_type not in ('ament_cmake', 'cmake') and
                not self.build_context.run_doxygen):
            logger.info(
                f"The package build type is not 'ament_cmake' or 'cmake', hence the"
                f"'{self.builder_type}' builder was not configured")
```

rosdoc2/verbs/build/builders/doxygen_builder.py
```diff
@@ -149,6 +158,14 @@ def __init__(self, builder_name, builder_entry_dictionary, build_context):
             logger.info(f"Using user specified Doxyfile at '{self.doxyfile}'.")
 
     def build(self, *, doc_build_folder, output_staging_directory):
+        # If the build type is not 'ament_cmake/cmake', there is no reason to run doxygen.
+        if (self.build_context.build_type not in ['ament_cmake', 'cmake'] and
+                not self.build_context.run_doxygen):
+            logger.info(
+                f"The package build type is not 'ament_cmake/cmake', hence the"
+                f"'{self.builder_type}' builder is not invoked")
```
clalancette, line 166:
```suggestion
        # If the build type is not 'ament_cmake/cmake', there is no reason to run doxygen.
        if (self.build_context.build_type not in ('ament_cmake', 'cmake') and
                not self.build_context.run_doxygen):
            logger.info(
                f"The package build type is not 'ament_cmake' or 'cmake', hence the"
                f"'{self.builder_type}' builder was not invoked")
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -53,49 +69,50 @@ def ensure_global(name, default):
...
-        # Maps markdown files to the "md" lexer, and not the "markdown" lexer
-        # Pygments registers "md" as a valid markdown lexer, and not "markdown"
-        "lexerMapping": {{r".*\.(md|markdown)$": "md",}},
-        # This mapping will work when `exhale` supports `:doxygenpage:` directives
-        # Check https://github.com/svenevs/exhale/issues/111
-        # TODO(aprotyas): Uncomment the mapping below once the above issue is resolved.
-        # "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
-        #     lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
-    }})
+if '{build_type}' != 'ament_python' or {run_doxygen}:
```
clalancette, line 72:
```suggestion
if '{build_type}' in ('ament_cmake', 'cmake') or {run_doxygen}:
```

(this is so that if we add a new build type support, this will still do the correct thing)

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -320,7 +334,9 @@ def __init__(self, builder_name, builder_entry_dictionary, build_context):
 
     def build(self, *, doc_build_folder, output_staging_directory):
         # Check that doxygen_xml_directory exists relative to output staging, if specified.
-        if self.doxygen_xml_directory is not None:
+        if (self.doxygen_xml_directory is not None and
+                (self.build_context.build_type != 'ament_python' or
```
clalancette, line 338:
```suggestion
                (self.build_context.build_type in ('ament_cmake', 'cmake') or
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -372,6 +388,50 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
+                            package_xml_directory,
+                            package_list[package_list.index(self.build_context.package.name)])
+                except ValueError:
+                    package_src_directory = ''
+            if not package_src_directory:
+                msg = \
+                'Could not locate source directory to invoke sphinx-apidoc in. ' \
+                'If this is package does not have a standard Python package layout, '\
+                'please specify the Python source in `rosdoc2.yaml`.'
+                raise RuntimeError(msg)
```
clalancette, line 417:
```suggestion
                raise RuntimeError('Could not locate source directory to invoke sphinx-apidoc in. '
                                   'If this is package does not have a standard Python package layout, '
                                   'please specify the Python source in `rosdoc2.yaml`.')
```

clalancette (dismissed):
I've left a few minor things to fix up.  Mostly I want to make sure that this will continue to work if a new documentation type is added.

Separately, I tried running this against 'launch', and it did indeed work (which is great).  But I got quite a few errors when running, of the form:

```
WARNING: autodoc: failed to import module 'substitutions.this_launch_file' from module 'launch'; the following exception was raised:
No module named 'launch'
```

rosdoc2/verbs/build/inspect_package_for_settings.py
```diff
@@ -44,6 +44,23 @@
...
+    ## This setting, if true, attempts to run `doxygen` and the `breathe`/`exhale`
+    ## extensions to `sphinx` regardless of build type. This is most useful if the
+    ## user would like to generate C/C++ API documentation for a package that is not
+    ## of the `ament_cmake/cmake` build type.
+    run_doxygen: false
+
+    ## This setting, if true, attempts to run `sphinx-apidoc` regardless of build
+    ## type. This is most useful if the user would like to generate Python API
+    ## documentation for a package that is not of the `ament_python` build type.
+    run_sphinx_apidoc: false
```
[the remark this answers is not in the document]
aprotyas: Addressed in 146513e and ef8840e respectively.

rosdoc2/verbs/build/builders/doxygen_builder.py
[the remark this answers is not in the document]
aprotyas: As it stands, the `ament_cmake` and `cmake` build types behave identically in `rosdoc2`. One example:

```py
# If the build type is not `ament_cmake/cmake`, there is no reason
# to create a doxygen builder.
if (self.build_context.build_type not in ('ament_cmake', 'cmake') and
```

However, the structure as-is allows different behaviors easily. Should I resolve this conversation?

rosdoc2/verbs/build/builders/sphinx_builder.py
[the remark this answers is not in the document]
wjwwood: My concern is that a user explicitly turns exhale on (for example) with `rosdoc2_settings['enable_exhale'] = True` but it has no effect because they don't satisfy `'{build_type}' in ('ament_cmake', 'cmake') or {always_run_doxygen}`.

I think the behavior is that we always check the value of `enable_exhale` (for example), but we choose the default (currently always `True`) based on the build type and/or doxygen setting. If the user sets it to `True` manually (overriding the default) and doxygen is not being run we can throw a warning or error for that case.

rosdoc2/verbs/build/builders/doxygen_builder.py
```diff
@@ -77,6 +77,15 @@ def __init__(self, builder_name, builder_entry_dictionary, build_context):
...
         assert self.builder_type == 'doxygen'
 
+        # If the build type is not `ament_cmake/cmake`, there is no reason
+        # to create a doxygen builder.
+        if (self.build_context.build_type not in ('ament_cmake', 'cmake') and
+                not self.build_context.always_run_doxygen):
+            logger.info(
+                f"The package build type is not 'ament_cmake' or 'cmake', hence the"
+                f"'{self.builder_type}' builder was not configured")
+            return None
```
wjwwood, line 87: nitpick: I prefer this style of wrapping long conditionals so you never indent more than once between two consecutive lines but also avoid indentation ambiguity:

```suggestion
        if (
            self.build_context.build_type not in ('ament_cmake', 'cmake') and
            not self.build_context.always_run_doxygen
        ):
            logger.info(
                f"The package build type is not 'ament_cmake' or 'cmake', hence the"
                f"'{self.builder_type}' builder was not configured")
            return None
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -24,6 +25,21 @@
 
 logger = logging.getLogger('rosdoc2')
 
+def generate_package_toc_entry(*, build_context) -> str:
+    build_type = build_context.build_type
+    always_run_doxygen = build_context.always_run_doxygen
+    always_run_sphinx_apidoc = build_context.always_run_sphinx_apidoc
+    toc_entry_py = f'\n   {build_context.package.name} Python API <modules>'
+    toc_entry_cpp = '\n   api/library_root\n   Full C/C++ API <api/unabridged_api>\n   File structure <api/unabridged_orphan>'
```
wjwwood, line 33: nitpick: this line looks too long

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -320,7 +334,9 @@ def __init__(self, builder_name, builder_entry_dictionary, build_context):
 
     def build(self, *, doc_build_folder, output_staging_directory):
         # Check that doxygen_xml_directory exists relative to output staging, if specified.
-        if self.doxygen_xml_directory is not None:
+        if (self.doxygen_xml_directory is not None and
+                (self.build_context.build_type in ('ament_cmake', 'cmake') or
+                 self.build_context.always_run_doxygen)):
             self.doxygen_xml_directory = \
                 os.path.join(output_staging_directory, self.doxygen_xml_directory)
```
wjwwood, line 341: nitpick: similar suggestion to make it clearer to read:

```suggestion
        if (
            self.doxygen_xml_directory is not None and
            (
                self.build_context.build_type in ('ament_cmake', 'cmake') or
                self.build_context.always_run_doxygen
            )
        ):
            self.doxygen_xml_directory = \
                os.path.join(output_staging_directory, self.doxygen_xml_directory)
```

This one also avoids indentations that are not a multi of four spaces.

aprotyas:
On the parallel discussion about running against the 'launch' package. I just tried to do so locally, and it did work. I too got a bunch of errors but nothing along the lines of "failed to import module". The errors that I'm seeing are more problems in `.rst` markup within `launch` and duplicate references to some symbols, here's a snippet of the warnings:

<details>
<summary>Short snippet of the warnings from running <code>rosdoc2</code> on <code>launch</code></summary>

```
/home/abrar/ros2_rolling/build/launch/launch/event_handlers/event_named.py:docstring of launch.event_handlers.event_named:1: WARNING: duplicate object description of launch.event_handlers.event_named, other instance in launch.event_handlers, use :noindex: for one of them
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.extract_type:12: WARNING: Unexpected indentation.
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.is_substitution:6: WARNING: Unexpected indentation.
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.is_typing_list:4: WARNING: Unexpected section title.

Examples
--------
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.normalize_typed_substitution:7: WARNING: Title underline too short.

Example:
-------
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.normalize_typed_substitution:7: WARNING: Unexpected section title.

Example:
-------
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.normalize_typed_substitution:10: WARNING: Unexpected indentation.
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.normalize_typed_substitution:8: WARNING: Inline literal start-string without end-string.
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.normalize_typed_substitution:8: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.normalize_typed_substitution:17: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.normalize_typed_substitution:17: WARNING: Inline literal start-string without end-string.
/home/abrar/ros2_rolling/build/launch/launch/utilities/type_utils.py:docstring of launch.utilities.type_utils.normalize_typed_substitution:17: WARNING: Inline interpreted text or phrase reference start-string without end-string.
looking for now-outdated files... none found
pickling environment... done
checking consistency... /home/abrar/Documents/open_robotics/code/ros2_humble/src/launch/launch/doc/source/modules.rst: WARNING: document isn't included in any toctree
done
preparing documents... done
writing output... [100%] modules                                                               
/home/abrar/ros2_rolling/build/launch/launch/launch_description_sources/any_launch_file_utilities.py:docstring of launch.launch_description_sources.any_launch_file_utilities.get_launch_description_from_any_launch_file:: WARNING: more than one target found for cross-reference 'InvalidLaunchFileError': launch.InvalidLaunchFileError, launch.invalid_launch_file_error.InvalidLaunchFileError, launch.launch_description_sources.InvalidLaunchFileError
```

</details>

Regardless, when I was working on this PR I did consider multiple ways to mitigate errors like "can't find module", because `sphinx-apidoc` requires all modules to be in the system path. I settled on invoking `sphinx-apidoc` with the `-a` flag, which [`appends module_path to sys.path`](https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html#cmdoption-sphinx-apidoc-a). I'm not sure why you got those warnings @clalancette. Any clues/thoughts?

rosdoc2/verbs/build/builders/sphinx_builder.py
wjwwood: Also, don't be afraid to use more variables for intermediate values, e.g. the second half of the condition could be `should_run_doxygen` which is calculated on a previous line.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -372,6 +388,48 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
+            package_xml_directory = os.path.dirname(self.build_context.package.filename)
+            # If 'python_source' is specified, construct 'package_src_directory' from it
+            if self.build_context.python_source is not None:
+                package_src_directory = \
+                    os.path.join(
+                        package_xml_directory,
+                        self.build_context.python_source)
+            # If not provided, try to find the package source direcotry
+            else:
+                package_list = setuptools.find_packages(where=package_xml_directory)
```
wjwwood, line 404: I'll be interested to see what happens here when we scale this out to more of our python packages. In the past when using `setuptools.find_packages()` in my `setup.py` files I've had to exclude test directories and stuff like that.

rosdoc2/verbs/build/parse_rosdoc2_yaml.py
```diff
@@ -57,6 +57,17 @@ def parse_rosdoc2_yaml(yaml_string, build_context):
...
+        build_context.add_python_source(python_source=settings_dict['python_source'])
+
+    if 'always_run_doxygen' in settings_dict:
+        build_context.add_always_run_doxygen(
+            always_run_doxygen=settings_dict['always_run_doxygen'])
+
+    if 'always_run_sphinx_apidoc' in settings_dict:
+        build_context.add_always_run_sphinx_apidoc(
+            always_run_sphinx_apidoc=settings_dict['always_run_sphinx_apidoc'])
+
```
wjwwood, line 70: This pattern is weird to me. There are places where you unconditionally do things like `build_context.always_run_doxygen`, but the way this is set up if the matching case here were false, that code would result in an error because the build context will have no member `always_run_doxygen`.

Wouldn't it be better to give default values (maybe just `None`) to these members in the constructor of the build context and just set them here if they show up in the settings dict?

Maybe there's a reason you're doing it this way, but it seems like we're misusing the build context as a class.

wjwwood:
> One option to mitigate this is through manually calling sys.path.insert(0, os.path.abspath('path/to/module')) for all modules returned by a setuptools.find_package('path/to/module'). Thoughts?

I was going to suggest this a while ago but got caught up in a review too.

It's in the sphinx-autodoc examples that you may need to add the local module directories to the `sys.path` for autodoc to work.

rosdoc2/verbs/build/builders/doxygen_builder.py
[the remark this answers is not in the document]
wjwwood: The only important difference is that with `ament_cmake` we can more reasonably assume a particular layout of the files in the project. You find much more variety in plain cmake packages.

That's not really an issue I guess though since we can't fully assume anything in either case. So we can lump `cmake` in with `ament_cmake`.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -53,49 +69,50 @@ def ensure_global(name, default):
...
-        # Maps markdown files to the "md" lexer, and not the "markdown" lexer
-        # Pygments registers "md" as a valid markdown lexer, and not "markdown"
-        "lexerMapping": {{r".*\.(md|markdown)$": "md",}},
-        # This mapping will work when `exhale` supports `:doxygenpage:` directives
-        # Check https://github.com/svenevs/exhale/issues/111
-        # TODO(aprotyas): Uncomment the mapping below once the above issue is resolved.
-        # "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
-        #     lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
-    }})
+if '{build_type}' in ('ament_cmake', 'cmake') or {always_run_doxygen}:
```
wjwwood, line 72: nitpick: consider reading this generated file, you'd see something like `if 'foo' in ('ament_cmake', 'cmake') or False:`. This could be confusing, so consider doing something like:

```suggestion
build_type = '{build_type}'
always_run_doxygen = {always_run_doxygen}
if build_type in ('ament_cmake', 'cmake') or always_run_doxygen:
```

rosdoc2/verbs/build/builders/sphinx_builder.py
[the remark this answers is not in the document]
aprotyas: Okay, I see what you mean now. I didn't consider that approach because I thought there was no reason to turn exhale on unless it's an `ament_cmake`/`cmake` build type, but in hindsight you could generate doxygen tags for many types of projects. I will address this in a commit.

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Yes, this is the first thing I've noticed too. One immediate idea is to do a wildcard search such as `[pkg-name]*`, and only add those modules to the path.

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Addressed in 8d61f5b.

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Addressed in 1d7e75d.

rosdoc2/verbs/build/parse_rosdoc2_yaml.py
aprotyas: Hmm, my bad. I conflated the absence of a class member (i.e. a type does **not** have an attribute) with that attribute just having a `None` value. I will address that with default `None` initialization in the build_context constructor.

---

About the use of the build context in this manner: I am using it in this way to propagate information from the build invocation through to each of the builders (Doxygen, Sphinx). I wasn't sure of another way to exhibit this behavior, and I believe settings like `always_run_doxygen` do fall under the scope of build context. Having said that, we could very well just introduce a new object type to propagate this information throughout `rosdoc2`, or maybe just introduce/read these settings elsewhere. I'm not convinced they're better than the current use of the build context as a class though. Thoughts?

aprotyas:
Here's what's changed since the latest round of feedback:

- Formatting changes within the builder classes: more legible indentation of if-conditions, limiting line length, etc. (3379546, 8d61f5b, 1d7e75d, 51e11ed)
- Made the generated `conf.py` file less "cryptic" for the user, i.e. format mapped template variables into local variables before using them. (f2415cf)
- Initialized some `rosdoc2` settings being propagated through the `BuildContext` class with reasonable default values. (0642f05)
- Inserted the package source directory into `sys.path` and provided `sphinx-autodoc` a list of `exec_dependency`s it could patch were they not present in `sys.path`. More about this below. (cecccee, eef02f1)
- Fixed up logic to check that the correct package source directory was identified. (final patch in 5d84071)

---

> > One option to mitigate this is through manually calling sys.path.insert(0, os.path.abspath('path/to/module')) for all modules returned by a setuptools.find_package('path/to/module'). Thoughts?
> 
> I was going to suggest this a while ago but got caught up in a review too.
> 
> It's in the sphinx-autodoc examples that you may need to add the local module directories to the `sys.path` for autodoc to work.

Yes, you're right. `sphinx-autodoc` actually has to import modules and hence requires the module source directories in `sys.path`.

As I said in https://github.com/ros-infrastructure/rosdoc2/pull/28#issuecomment-897026854, I wrongly assumed the `-a` flag used to invoke `sphinx-apidoc` would insert said directories into `sys.path`. There are two problems here:

- That flag only works if you're using `sphinx-apidoc` to generate a full `sphinx` project, which we are not.
- We cannot actually insert these directories into `sys.path` at the `SphinxBuilder` class layer, because of the scope of this change. When `sphinx-build` is invoked - and consequently, when `sphinx-autodoc` runs - it does not seem to find the modifications made to `sys.path`.

The solution I've settled on is:

- Provide the package source directory as a template variable to the `rosdoc2_wrapping_conf_py_template`. In this template, if the `enable_autodoc` setting is True, the package source directory's parent is inserted to `sys.path`. i.e. `sys.path.insert(0, os.path.dirname('{package_src_directory}')`.
- Parallel to the above, provide the list of `exec_depend`s of the package being processed as a template variable to the `rosdoc2_wrapping_conf_py_template`. Again, if `enable_autodoc` is True, this list of `exec_depend`s is assigned to `autodoc_mock_imports`. Since `autodoc` imports every module in a file - it will invariably attempt to import that package's `exec_depend`s too. Since it is possible for those dependencies to not be in `sys.path`, `autodoc` will mock any modules that it fails to import. This allows `autodoc` to build till completion, basically.

Thoughts about the solution? I could be wrong about my assessment of `We cannot actually insert these directories into `sys.path` at the `SphinxBuilder` class layer`, but I couldn't get anything to work that way. Let me know if this solution needs modification.

---

I believe the PR has addressed most of the feedback provided so far. Thoughts?

clalancette:
>     * Provide the package source directory as a template variable to the `rosdoc2_wrapping_conf_py_template`. In this template, if the `enable_autodoc` setting is True, the package source directory's parent is inserted to `sys.path`. i.e. `sys.path.insert(0, os.path.dirname('{package_src_directory}')`.

One important thing we are going to have to keep in mind with this change is the buildfarm.  In particular, the current rosdoc2 jobs do not install dependencies or source the ROS environment while building the documentation.  In order for this to work, we are going to have to change the buildfarm to do both of those things, otherwise the whole `spinx-autodoc` thing isn't going to be able to find the pieces of the environment that it expects.

aprotyas:
I don't think anything will have to be changed in the buildfarm for `sphinx-autodoc` to work though. The dependencies will all be patched by `sphinx-autodoc`, and only the source directory of the package in question gets appended (temporarily) to `sys.path`.

wjwwood:
Yeah, my hope is that autodoc just needs to include python modules that are local to the package, but I haven't tested that.

---

All the rest of what you mentioned makes sense to me @aprotyas. I'll do another code review.

rosdoc2/verbs/build/builders/doxygen_builder.py
```diff
@@ -77,6 +77,17 @@ def __init__(self, builder_name, builder_entry_dictionary, build_context):
...
+        # If the build type is not `ament_cmake/cmake`, there is no reason
+        # to create a doxygen builder.
+        if (
+            self.build_context.build_type not in ('ament_cmake', 'cmake') and
+            not self.build_context.always_run_doxygen
+        ):
+            logger.info(
+                f"The package build type is not 'ament_cmake' or 'cmake', hence the"
+                f"'{self.builder_type}' builder was not configured")
+            return None
```
wjwwood, line 89: `return None` is kind of odd here since it's a constructor. Maybe just `return`, which is the same thing really, but at least it implies there's no return type that matters here.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -15,21 +15,41 @@
...
 logger = logging.getLogger('rosdoc2')
 
+def generate_package_toc_entry(*, build_context) -> str:
+    build_type = build_context.build_type
+    always_run_doxygen = build_context.always_run_doxygen
+    always_run_sphinx_apidoc = build_context.always_run_sphinx_apidoc
+    toc_entry_py = f'\n   {build_context.package.name} Python API <modules>'
+    toc_entry_cpp = \
+        '\n   api/library_root\n   Full C/C++ API <api/unabridged_api> \
+        \n   File structure <api/unabridged_orphan>'
```
wjwwood, line 36: I'd recommend using a multi-line string for this, it's a bit easier to read.

```suggestion
    toc_entry_cpp = """
   api/library_root
   Full C/C++ API <api/unabridged_api>
   File structure <api/unabridged_orphan>"""
```

Also did you mean for it to be indented by three spaces inside the string? Maybe that's right for the rst content, but I just wanted to check, since that's kind of odd.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -366,10 +405,55 @@ def build(self, *, doc_build_folder, output_staging_directory):
             if package_name != self.build_context.package.name
         ]
 
+        # If the package has build type `ament_python`, or if the user configured
+        # to run `sphinx-apidoc`, then invoke `sphinx-apidoc` before building
+        if (
+            self.build_context.build_type == 'ament_python' or
+                self.build_context.always_run_sphinx_apidoc
+        ):
```
wjwwood, line 413: nitpick:

```suggestion
        if (
            self.build_context.build_type == 'ament_python' or
            self.build_context.always_run_sphinx_apidoc
        ):
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -366,10 +405,55 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
+            # If not provided, try to find the package source direcotry
+            else:
+                package_list = setuptools.find_packages(where=package_xml_directory)
+                try:
+                    package_src_directory = \
+                        os.path.join(
+                            package_xml_directory,
+                            package_list[package_list.index(self.build_context.package.name)])
+                except ValueError:
+                    package_src_directory = None
```
wjwwood, line 430: Ok, I see how this address the "over matching" of `setuptools.find_packages()` 👍.

However, the try-except is kind of odd. I think it's ok to leave as-is, but an explicit `if ... in package_list` might be easier to understand. At first glance it's not clear where the `ValueError` is coming from. Maybe something like:

```suggestion
                if self.build_context.package.name in package_list:
                    package_src_directory = os.path.join(
                        package_xml_directory,
                        self.build_context.package.name)
                else:
                    package_src_directory = None
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -366,10 +405,55 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
+
+            if not package_src_directory or not os.path.isdir(package_src_directory):
+                raise RuntimeError('Could not locate source directory to invoke sphinx-apidoc in. '
+                                   'If this is package does not have a standard Python package layout, '
+                                   'please specify the Python source in `rosdoc2.yaml`.')
```
wjwwood, line 435: nitpick: again try to avoid indents that are more than four from the previous line, or are not a multiple of four, and also we tend to use `'` inside messages to quote file names, etc...:

```suggestion
                raise RuntimeError(
                    "Could not locate source directory to invoke sphinx-apidoc in. "
                    "If this is package does not have a standard Python package layout, "
                    "please specify the Python source in 'rosdoc2.yaml'.")
```

rosdoc2/verbs/build/parse_rosdoc2_yaml.py
wjwwood: I think it's fine to pass additional information via the build context in this way.

rosdoc2/verbs/build/parse_rosdoc2_yaml.py
```diff
@@ -57,6 +57,17 @@ def parse_rosdoc2_yaml(yaml_string, build_context):
...
+    if 'python_source' in settings_dict:
+        build_context.add_python_source(python_source=settings_dict['python_source'])
+
+    if 'always_run_doxygen' in settings_dict:
+        build_context.add_always_run_doxygen(
+            always_run_doxygen=settings_dict['always_run_doxygen'])
+
+    if 'always_run_sphinx_apidoc' in settings_dict:
+        build_context.add_always_run_sphinx_apidoc(
+            always_run_sphinx_apidoc=settings_dict['always_run_sphinx_apidoc'])
```
wjwwood, line 69: I still think this is overkill, I don't think you need the `build_context.add_*` methods, e.g.:

```suggestion
    build_context.python_source = settings_dict.get('python_source', None)
    build_context.always_run_doxygen = settings_dict.get('always_run_doxygen', None)
    build_context.always_run_sphinx_apidoc = settings_dict.get('always_run_sphinx_apidoc', None)
```

And then remove the `add_*` methods from `BuildContext`.

ivanpauno:
Related to my last comment, the [exec()](https://github.com/ros-infrastructure/rosdoc2/blob/81f94745eb058634af1ef6d52631217783ce9a16/rosdoc2/verbs/build/builders/sphinx_builder.py#L56) line to include the user `conf.py` seems to be problematic if the original file had some relative files (depending where you run rosdoc2).

e.g. if you do things like https://docs.readthedocs.io/en/stable/guides/adding-custom-css.html.

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-08-19/22008/1

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-9-16/22372/1

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-10-28/22947/1

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-11-18/23209/1

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -366,10 +409,55 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
         ]
 
+        # If the package has build type `ament_python`, or if the user configured
+        # to run `sphinx-apidoc`, then invoke `sphinx-apidoc` before building
+        if (
+            self.build_context.build_type == 'ament_python' or
+            self.build_context.always_run_sphinx_apidoc
+        ):
+            package_xml_directory = os.path.dirname(self.build_context.package.filename)
+            # If 'python_source' is specified, construct 'package_src_directory' from it
```
rkent, line 419: This section that assigns a value to package_src_directory is behind an if statement, so it remains undefined if the 'if' does not get executed.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -366,10 +409,55 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
+            if completed_process.returncode == 0:
+                logger.info(msg)
+            else:
+                raise RuntimeError(msg)
+
         # Setup rosdoc2 Sphinx file which will include and extend the one in `sourcedir`.
         self.generate_wrapping_rosdoc2_sphinx_project_into_directory(
             doc_build_folder,
             sourcedir,
+            package_src_directory,
```
rkent, line 460: Here is where package_src_directory is used, outside of the if statement. This is also where my demo package fails "local variable 'package_src_directory' referenced before assignment"

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -366,10 +409,55 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
+            cmd = [
+                'sphinx-apidoc',
+                '-o', os.path.relpath(sourcedir, start=doc_build_folder),
+                '-e',  # Document each module in its own page.
+                os.path.abspath(package_src_directory),
+            ]
+            logger.info(
+                f"Running sphinx-apidoc: '{' '.join(cmd)}' in '{doc_build_folder}'"
+            )
+            completed_process = subprocess.run(cmd, cwd=doc_build_folder)
```
rkent, line 449: On first run, the variable doc_build_folder has been created with the correct path, but that path has not actually been created yet. The creation occurs later in generate_wrapping_rosdoc2_sphinx_project_into_directory. You either need to create the directory, or move generate_wrapping_rosdoc2_sphinx_project_into_directory to occur prior to this call.

I'm not really sure what the etiquette is in the ROS world around what we used to call "drive-by reviews" in the Mozilla world. Hopefully this is useful and not inappropriate. I have been trying to setup a mock ros2 package with all of the bells and whistles, including various rosdoc2 setups, so I am encountering these issues.

clalancette: > I'm not really sure what the etiquette is in the ROS world around what we used to call "drive-by reviews"

Very much appreciated!  This PR has been on our long-term list to test and get in, but we just haven't gotten to it yet.  We are hoping to soon, so pointing out these problems is very helpful, thanks.

rosdoc2/verbs/build/builders/sphinx_builder.py
rkent: Sorry to trickle in these comments like this. As written this did not work fort me (after I fixed other issues reported earlier.) Here is what works:

        # Setup rosdoc2 Sphinx file which will include and extend the one in `sourcedir`.
        self.generate_wrapping_rosdoc2_sphinx_project_into_directory(
            doc_build_folder,
            sourcedir,
            #package_src_directory,
            os.path.abspath(package_src_directory),
            intersphinx_mapping_extensions)

aprotyas:
Lest this PR is mistook as stale, it's still on my radar and I plan to pick it back up in a couple of weeks once my semester is complete. Thanks for the additional feedback @rkent!

rkent:
Here are possible fixes for my comments, basically moving some of the setup earlier in the code: https://github.com/rkent/rosdoc2/tree/rkjfixes

rkent:
It's very hard to make progress in rosdoc2 until this PR is landed. I've done a rebase to main so I could continue with my demo in [my fork](https://github.com/rkent/rosdoc2). What can I do to help move this forward?

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Thank you for the fix! Addressed in https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/45b8e08a534df0c8bc9003acc5db32e9da92f374.

aprotyas:
> It's very hard to make progress in rosdoc2 until this PR is landed. I've done a rebase to main so I could continue with my demo in [my fork](https://github.com/rkent/rosdoc2). What can I do to help move this forward?

Thanks for the nudge! I've cherry-picked your commit from https://github.com/rkent/rosdoc2/tree/rkjfixes to resolve issues with referencing paths before accessing them.

aprotyas:
Trying to get this PR rolling again.

Updates:
- I've addressed logic errors where paths (and variables) were accessed before creation by cherry-picking @rkent's suggested patch on this branch. (https://github.com/ros-infrastructure/rosdoc2/commit/45b8e08a534df0c8bc9003acc5db32e9da92f374)
- Different index/TOC entries are generated based on the build type. Notably, the `{package_name} generated/index` entry only works for packages where Doxygen is invoked. (https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/fa97f79c93e7000d0587da829886041b234329ea)
- The PR has green CI following some linter fixes. (https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/903b81abbae9a2c1beaa5589ad5a65dbb2f1ca81 and #43)
- The PR has been rebased on the main branch. There are no merge conflicts now. Unfortunately I also squashed the commit history in the process... my apologies.

Outstanding concerns:

- You can still see different behavior based on where `rosdoc2` is run. @ivanpauno pointed out that if a package does anything with files local to its `conf.py` script, it may cause issues when it is processed with `exec`.
https://github.com/ros-infrastructure/rosdoc2/blob/81f94745eb058634af1ef6d52631217783ce9a16/rosdoc2/verbs/build/builders/sphinx_builder.py#L56
- #34 
- #42 

At this point, I'm putting out a request for reviews/comments/testing. FYI @clalancette @wjwwood @ivanpauno @hidmic @rkent

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-january-20th-2022/23986/1

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -48,30 +71,51 @@ def ensure_global(name, default):
...
+
+if rosdoc2_settings.get('enable_breathe', is_doxygen_invoked):
     # Configure Breathe.
     # Breathe ingests the XML output from Doxygen and makes it accessible from Sphinx.
     print('[rosdoc2] enabling breathe', file=sys.stderr)
+    # First check that doxygen would have been run
+    if not is_doxygen_invoked:
+        raise RuntimeError(
+            "Cannot enable the 'breathe' extension if 'doxygen' is not invoked."
+            "Please enable 'always_run_doxygen' if the package is not an"
```
clalancette, line 97:
```suggestion
            "Cannot enable the 'breathe' extension if 'doxygen' is not invoked. "
            "Please enable 'always_run_doxygen' if the package is not an "
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -48,30 +71,51 @@ def ensure_global(name, default):
...
     # Exhale uses the output of Doxygen and Breathe to create easier to browse pages
     # for classes and functions documented with Doxygen.
     # This is similar to the class hierarchies and namespace listing provided by
     # Doxygen out of the box.
     print('[rosdoc2] enabling exhale', file=sys.stderr)
+    # First check that doxygen would have been run
+    if not is_doxygen_invoked:
+        raise RuntimeError(
+            "Cannot enable the 'breathe' extension if 'doxygen' is not invoked."
+            "Please enable 'always_run_doxygen' if the package is not an"
```
clalancette, line 117:
```suggestion
            "Cannot enable the 'breathe' extension if 'doxygen' is not invoked. "
            "Please enable 'always_run_doxygen' if the package is not an "
```

rosdoc2/verbs/build/builders/doxygen_builder.py
```diff
@@ -152,6 +163,16 @@ def __init__(self, builder_name, builder_entry_dictionary, build_context):
...
     def build(self, *, doc_build_folder, output_staging_directory):
         """Actually do the build."""
+        # If the build type is not 'ament_cmake/cmake', there is no reason to run doxygen.
+        if (
+            self.build_context.build_type not in
+            ('ament_cmake', 'cmake') and not self.build_context.always_run_doxygen
+        ):
+            logger.info(
+                f"The package build type is not 'ament_cmake' or 'cmake', hence the"
+                f"'{self.builder_type}' builder was not invoked")
```
clalancette, line 173:
```suggestion
            logger.debug(
                f"The package build type is not 'ament_cmake' or 'cmake', hence the "
                f"'{self.builder_type}' builder was not invoked")
```

rosdoc2/verbs/build/builders/doxygen_builder.py
```diff
@@ -79,6 +79,17 @@ def __init__(self, builder_name, builder_entry_dictionary, build_context):
...
 
+        # If the build type is not `ament_cmake/cmake`, there is no reason
+        # to create a doxygen builder.
+        if (
+            self.build_context.build_type not in
+            ('ament_cmake', 'cmake') and not self.build_context.always_run_doxygen
+        ):
+            logger.info(
+                f"The package build type is not 'ament_cmake' or 'cmake', hence the"
+                f"'{self.builder_type}' builder was not configured")
```
clalancette, line 90:
```suggestion
            logger.debug(
                f"The package build type is not 'ament_cmake' or 'cmake', hence the "
                f"'{self.builder_type}' builder was not configured")
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -379,12 +424,61 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
+                '-e',  # Document each module in its own page.
+                package_src_directory,
+            ]
+            logger.info(
+                f"Running sphinx-apidoc: '{' '.join(cmd)}' in '{doc_build_folder}'"
+            )
+            completed_process = subprocess.run(cmd, cwd=doc_build_folder)
+            msg = f"sphinx-apidoc exited with return code '{completed_process.returncode}'"
+            if completed_process.returncode == 0:
+                logger.info(msg)
```
clalancette, line 478:
```suggestion
                logger.debug(msg)
```

clalancette (commented):
@aprotyas Sorry for the long delay here.  I finally found some time to look at this.

Overall, fantastic work.  Thanks so much for pushing this forward.  A few overall remarks:

First, I tested this on several Python packages in the ROS 2 core.  I was able to successfully generate documentation for some of them, but not others.  Successful generation was in packages like `launch`, `launch_xml`, `launch_testing`, etc.  Failures were in cases like `sros2` and `launch_ros`, which generally had failures like the following:

```
WARNING: autodoc: failed to import module 'utilities.track_node_names' from module 'launch_ros'; the following exception was raised:
Traceback (most recent call last):
  File ".local/lib/python3.8/site-packages/sphinx/ext/autodoc/importer.py", line 70, in import_module
    return importlib.import_module(modname)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "ros2_ws/src/ros2/launch_ros/launch_ros/launch_ros/__init__.py", line 17, in <module>
    from . import actions
  File "ros2_ws/src/ros2/launch_ros/launch_ros/launch_ros/actions/__init__.py", line 17, in <module>
    from .composable_node_container import ComposableNodeContainer
  File "ros2_ws/src/ros2/launch_ros/launch_ros/launch_ros/actions/composable_node_container.py", line 27, in <module>
    from .node import Node
  File "ros2_ws/src/ros2/launch_ros/launch_ros/launch_ros/actions/node.py", line 48, in <module>
    from launch_ros.parameters_type import SomeParameters
  File "ros2_ws/src/ros2/launch_ros/launch_ros/launch_ros/parameters_type.py", line 50, in <module>
    SomeSubstitutionsType_types_tuple +
TypeError: unsupported operand type(s) for +: 'SomeSubstitutionsType_types_tuple' and 'tuple'
```

It would be great if you could try to look into those, and maybe suggest a way forward for those packages.  I've also left some (minor) comments inline.

Regardless, I'm actually somewhat inclined to take this PR more-or-less as-is, and then iterate to fix more things in the future.  This is at least a good part of the way there.

rosdoc2/verbs/build/builders/doxygen_builder.py
aprotyas: Addressed in https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/7d0b4fd4dff3ca0c92f8364ab39b47625f944dbf

rosdoc2/verbs/build/builders/doxygen_builder.py
aprotyas: Addressed in https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/7d0b4fd4dff3ca0c92f8364ab39b47625f944dbf

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Addressed in https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/7d0b4fd4dff3ca0c92f8364ab39b47625f944dbf

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Addressed in https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/fb0eab5d53133b0a0d71a59edb7213a36c6fb522

rosdoc2/verbs/build/builders/sphinx_builder.py
aprotyas: Addressed in https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/fb0eab5d53133b0a0d71a59edb7213a36c6fb522

aprotyas:
@clalancette I addressed the feedback in https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/7d0b4fd4dff3ca0c92f8364ab39b47625f944dbf and https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/fb0eab5d53133b0a0d71a59edb7213a36c6fb522. Thanks for the review.

---

> It would be great if you could try to look into those, and maybe suggest a way forward for those packages. I've also left some (minor) comments inline
#### The problem
1. `sphinx_apidoc` is invoked, which uses the `autodoc` extension for automatic generation of Sphinx rst sources. 
2. `autodoc` requires that all depended packages be importable at runtime. This is not ideal for document generation because users would need all exec dependencies installed and on their system path when generating documents.
3. To deal with the restriction in (2), imported modules can be mocked up by populating the [`autodoc_mock_imports`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports), from https://github.com/aprotyas/rosdoc2/blob/fb0eab5d53133b0a0d71a59edb7213a36c6fb522/rosdoc2/verbs/build/builders/sphinx_builder.py#L71-L78:
```py
if rosdoc2_settings.get('enable_autodoc', True):
    print('[rosdoc2] enabling autodoc', file=sys.stderr)
    extensions.append('sphinx.ext.autodoc')
    # Provide all runtime dependencies to be mocked up
    # Note: `autodoc` only mocks up those modules that it actually cannot locate in PATH
    autodoc_mock_imports = {exec_depends}
    # Add the package directory to PATH, so that `sphinx-autodoc` can import it
    sys.path.insert(0, os.path.dirname('{package_src_directory}'))
```
4. `autodoc` also imports the module to be documented. When doing so, it executes all side-effects said module has on an import: i.e. executes the current module's `__init__.py` file.
5. For both of the failed builds in `sros2` and `launch_ros`, importing the module executes code which uses some object that has already been mocked in (3). As such, `importlib.import_module` throws an exception since it doesn't know, say, that `SomeSubstitutionsType_types_tuple` and `tuple` can be added with the `+` operand.

I've verified that this is the case because NOT mocking `launch` in the case of `launch_ros` fixed this issue locally. But, this is not a real fix because it brings us back to the restrictions in (2).

#### The solution
`autodoc`'s documentation issues a warning to this effect too:

>Warning
>[autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#module-sphinx.ext.autodoc) imports the modules to be documented. If any modules have side effects on import, these will be executed by autodoc when sphinx-build is run.
>
>If you document scripts (as opposed to library modules), make sure their main routine is protected by a if `__name__ == '__main__'` condition.

But, I don't think the proposed solution works for the packages in question, given what the `__init__.py` file of these packages is trying to accomplish.

As it stands I don't have a reasonable way forward here. Not mocking dependencies is not a useful compromise and neither is protecting import routines by a `__name__ == '__main__'` condition. Any ideas?

wjwwood:
> autodoc requires that all depended packages be importable at runtime. This is not ideal for document generation because users would need all exec dependencies installed and on their system path when generating documents.

I mean, we need the "full build" doc jobs eventually (for C++ packages and message packages) where you would install dependencies and build the package before running documentation. So eventually we want that anyway.

> For both of the failed builds in sros2 and launch_ros, importing the module executes code which uses some object that has already been mocked in (3). As such, importlib.import_module throws an exception since it doesn't know, say, that SomeSubstitutionsType_types_tuple and tuple can be added with the + operand.

Can you give a concrete example of this? I can't think of any examples from memory where this would be an issue. It must some globals or something.

aprotyas:
> I mean, we need the "full build" doc jobs eventually (for C++ packages and message packages) where you would install dependencies and build the package before running documentation. So eventually we want that anyway.

Right. Thinking about this again, there's no need to mock dependencies if we hold the condition that all dependencies are built/installed and rosdoc2 is invoked after sourcing an appropriate underlay workspace.

In fact, I just tried building documentation for sros2 and launch_ros under the "full build" conditions, and they seem to work perfectly - discounting some sphinx tag errors.

---

> Can you give a concrete example of this? I can't think of any examples from memory where this would be an issue. It must some globals or something.

Sure. If you look at a section of sros2's `__init__.py` file:

```python
_xml_cache_path = urllib.parse.urljoin(
    'file:',
    urllib.request.pathname2url(
        os.path.join(
            ament_index_python.get_package_share_directory('sros2'),
            'xml_cache',
            'xhtml-cache.xml'
        )
    )
)
```

If the dependencies for sros2 are mocked up, then `os.path.join` does not actually get a path from `ament_index_python.get_package_share_directory` (since this method has been mocked). Well, if I ignore my (arbitrary) need to mock dependencies, this isn't a problem anymore.

---

Thank you for the suggestion @wjwwood. I think the way forward is to just not mock dependencies, instead requiring that all dependencies are built/installed and rosdoc2 is invoked after sourcing an appropriate underlay workspace. Should this expectation be documented somewhere?

wjwwood:
> Sure. If you look at a section of sros2's __init__.py file:

That's a great example of something that could either be deferred to the first time it is used, or put into a function, to avoid the issue. For example, it could be this instead:

```python
_xml_cache_path = None


def _get_xml_cache_path
    global _xml_cache_path
    _xml_cache_path = urllib.parse.urljoin(
        'file:',
        urllib.request.pathname2url(
            os.path.join(
                ament_index_python.get_package_share_directory('sros2'),
                'xml_cache',
                'xhtml-cache.xml'
            )
        )
    )
    return _xml_cache_path
```

So I think mocking is ok, but we just need to fix the issues that are coming up. Long term we could avoid that when we have full builds, but we don't have that right now. So I'd say add back the mocking and fix these issues maybe.

aprotyas:
> So I think mocking is ok, but we just need to fix the issues that are coming up. Long term we could avoid that when we have full builds, but we don't have that right now. So I'd say add back the mocking and fix these issues maybe.


I've thought of this approach, but this pushes the problem to users, i.e. if we can't build sros2/launch_ros/pkg_foo documentation with rosdoc2, let's see if there are ways to address that in sros2/launch_ros/pkg_foo. Each package in question needs to come up with a specific fix to its "avoid side effects because my dependencies may be mocked" issue. I don't know if we want to do that, especially because we can already afford full builds (they work for both ament_cmake and ament_python packages)

rkent:
As @clalancette said,  this PR really needs to get landed and issues dealt with in followup.

That being said, in the issue of mocking imports, I'm not sure what "# Note: `autodoc` only mocks up those modules that it actually cannot locate in PATH" means. The issue in the examples is `launch` and even if the user has sourced ROS2 so that `launch` can be imported (which is how I interpret as 'in PATH') that mock still occurs.

For a user who is running rosdoc2 in an environment where everything they need for their package is already installed, the issue can be fixed in rosdoc2 by importing the modules in conf.py. That is, modify `rosdoc2_wrapping_conf_py_template` to include something like this:
```python
    import importlib
    for item in autodoc_mock_imports:
        try:
            importlib.import_module(item)
        except:
            pass
```
Or alternatively just stop doing the mock.

This would allow rosdoc2 to run successfully locally, but then fail when run in an infrastructure environment that produces common documentation. You could always install modules when running rodoc2, but that could be always slow. Perhaps you could add a runtime option to rosdoc2 to install modules prior to running that would be used in the infrastructure setting.

Anyway this discussion belongs in a separate issue after this patch has landed.

aprotyas:
> I'm not sure what "# Note: `autodoc` only mocks up those modules that it actually cannot locate in PATH" means. The issue in the examples is `launch` and even if the user has sourced ROS2 so that `launch` can be imported (which is how I interpret as 'in PATH') that mock still occurs.

Your interpretation is correct, and my comment there is wrong - I retracted it in https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/56a79083ab4163b705872c76e08b7ddfa76687c6.

> Or alternatively just stop doing the mock.

I think this recommendation is reasonable for packages that can actually be imported. From https://github.com/ros-infrastructure/rosdoc2/pull/28/commits/56a79083ab4163b705872c76e08b7ddfa76687c6:

```python
    pkgs_to_mock = []
    import importlib
    for exec_depend in {exec_depends}:
        try:
            importlib.import_module(exec_depend)
        except:
            pkgs_to_mock.append(exec_depend)
    autodoc_mock_imports = pkgs_to_mock
```

I'm not fully clear on rosdoc2's expectations regarding the presence of package dependencies on the system, so I haven't dropped mocking outright. Note that the `autodoc` extension itself is still needed because some generated rst files contain autodoc directives.

Again, let's defer that conversation to after this PR has landed.

---

@clalancette I think this PR is now ready to land. I've tested that the package builds correctly with launch, launch_testing, launch_pytest, launch_xml, launch_yaml, launch_ros, ros2cli, ros2launch, sros2. I've also checked that rosdoc2 builds for C++ packages like rclcpp, rcpputils, rcl, rmw behave the same as before.

rkent:
What will it take to get this landed? I'd like to do some work on rosdoc2, but I really cannot make progress as long as this PR is outstanding.

aprotyas:
This PR is ready to land from my end too.

Yadunund (commented):
@clalancette I had a chance to test this PR against `launch_ros/ros2launch` which is an `ament_python` package and `rclpy` which is an `ament_cmake` package with python APIs to document. 

No problems with `ros2launch`.
image

With `rclpy`, I had to add a [rosdoc2.yaml](https://github.com/ros2/rclpy/compare/yadu/add_rosdoc2_yaml) along with a `<rosdoc2>` export tag in the package.xml. The yaml explictly disables `doxygen` generation which will execute by default else. However another issue exists with such packages where the `sphinx_builder` script will assume doxygen was invoked given the logic introduced [here](https://github.com/aprotyas/rosdoc2/blob/0016ae498ced22a5aeddabee4da94209ca26f456/rosdoc2/verbs/build/builders/sphinx_builder.py#L91-L93). As a result, it attempts to invoke `breathe` and `exhale` which will result in an error. That `rosdoc2_settings` dict is defined in the [same script](https://github.com/aprotyas/rosdoc2/blob/0016ae498ced22a5aeddabee4da94209ca26f456/rosdoc2/verbs/build/builders/sphinx_builder.py#L266-L273) and it seems like the default entries do not get overwritten/updated by the key-values specified in the `rosdoc2.yaml`. So in the end to generate the doc for this pkg I had to explictly set `enable_breathe` and `enable_exhale` to `False` within the script. 
image

I think it should be possible to update that `rosdoc2_settings` dict with values from the `rosdoc2.yaml` either by injecting suitable entires into the [template_variables](https://github.com/aprotyas/rosdoc2/blob/0016ae498ced22a5aeddabee4da94209ca26f456/rosdoc2/verbs/build/builders/sphinx_builder.py#L600-L610) from the `BuildContext` or by updating the implementation of [generate_package_toc_entry](https://github.com/aprotyas/rosdoc2/blob/0016ae498ced22a5aeddabee4da94209ca26f456/rosdoc2/verbs/build/builders/sphinx_builder.py#L30).
I'm happy to work on this in a follow up PR if we want to merge this first.

rkent:
A few months ago, I did a bit of work on rosdoc2. My repo at https://github.com/rkent/rosdoc2 explored various approaches. I wanted to start to incorporate things in the main rosdoc2. But no progress could be made until the current patch is landed. Requests from several of us to get it landed went unanswered. So I lost interest, and have since moved on to other projects.

One of the last things that I did was a branch that added a basic unit test framework, rkent/add-tests, which was to be my suggestion of the minimum changes before suggesting some more radical changes. As part of that, I fixed some obvious issues that the tests revealed, including the issue in @clalancette's comment.

This commit seemed to make the issue go away:

https://github.com/rkent/rosdoc2/commit/3fb714120e594e80f58d8fe45658f369368fd13d

Essentially the fix that I proposed was to add this line to the default conf.py:

sys.path.insert(0, os.path.abspath(os.path.join('{package_src_directory}', '..')))

So, where to go from here? My preference would be to land @aprotyas's patch as-is, and fix issues in subsequent patches. There are other issues with the patch, but it is hard to address them with the current patch unlanded behind a 100 comment thread. It would be much easier to address them with smaller patches, backed up by a basic unit test framework like rkent/add-tests. IIRC rosdoc2 is hardly used in production (has that changed?) so the downsides of just getting it landed are minimal.

What this patch really needs is for someone with the power to review and land patches to either land this patch as-is, or specify the minimum needed along with a commitment to review and land in a timely fashion.

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-02-16/29927/1

Yadunund:
@rkent I can understand your frustration. Just want to point out that `rosdoc2` is actively used in production- we run document generation jobs on our build farms using this tool. Hence, the caution with merging big changes that may potentially result in build failures across hundreds of packages.

Having said that, i'm keen to get `rosdoc2` operational for python packages. I've created this [checklist ticket](https://github.com/ros-infrastructure/rosdoc2/issues/48) to delineate the steps needed for the same. One of the steps is to get your commit patched in so kindly keep an eye out for a PR request once we merge this PR in. There are other changes also needed as listed in the ticket.

My plan is to do one final check that the proposed changes do not break any documentation generation of `ament_cmake` packages before working together with the maintainers to get the proposed changes in. Stay tuned!

Yadunund (approved):
LGTM. Will be opening a follow-up PR with some other changes.

ros-discourse:

https://discourse.ros.org/t/growing-issue-with-ros-documentation/36075/48

PR #31, opened 2021-08-19
Pin to `breathe<=4.26.0`

aprotyas:
As discovered in https://github.com/svenevs/exhale/issues/98,
older versions of `breathe` do not throw unresolved function
name warnings and/or choke on template parameter lists.

Hence, temporarily pinning to 4.26.0.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ keywords = rosdoc2
 [options]
 python_requires = >=3.5
 install_requires =
-    breathe
+    breathe<=4.26.0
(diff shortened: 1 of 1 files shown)


aprotyas:
Additionally, this PR also requires pinning to a specific version of `sphinx` (i.e. `sphinx<3.5,>=3.0`). In offline discussions with @clalancette, we concluded that pinning both `breathe` and `sphinx` to specific older versions can be troublesome/impossible for users.
As such, I am closing this PR here.

PR #33, opened 2021-09-30
Revamp default Sphinx project

hidmic:
This patch:
- Modifies default index
- Uses Exhale generated index
- Changes Exhale configuration

Requires https://github.com/svenevs/exhale/pull/114.

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -78,23 +78,22 @@ if rosdoc2_settings.get('enable_exhale', True):
     from exhale import utils
     exhale_args.update({{
         # These arguments are required.
-        "containmentFolder": "{user_sourcedir}/api",
-        "rootFileName": "library_root.rst",
-        "rootFileTitle": "{package_name} API",
+        "containmentFolder": "{user_sourcedir}/generated",
+        "rootFileName": "index.rst",
         "doxygenStripFromPath": "..",
         # Suggested optional arguments.
         "createTreeView": True,
+        "fullToctreeMaxDepth": 1,
+        "unabridgedOrphanKinds": [],
+        "fullApiSubSectionTitle": "Reference",
         # TIP: if using the sphinx-bootstrap-theme, you need
         # "treeViewIsBootstrap": True,
         "exhaleExecutesDoxygen": False,
         # Maps markdown files to the "md" lexer, and not the "markdown" lexer
         # Pygments registers "md" as a valid markdown lexer, and not "markdown"
         "lexerMapping": {{r".*\.(md|markdown)$": "md",}},
-        # This mapping will work when `exhale` supports `:doxygenpage:` directives
-        # Check https://github.com/svenevs/exhale/issues/111
-        # TODO(aprotyas): Uncomment the mapping below once the above issue is resolved.
-        # "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
-        #     lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
+        "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
+            lambda kind: [":content-only:", ":no-link:"] if kind == "page" else []),
     }})
 
 if rosdoc2_settings.get('override_theme', True):
@@ -246,20 +245,13 @@ rosdoc2_settings = {{
 """
 
 index_rst_template = """\
-{package.name}
-{package_underline}
-
-{package.description}
-
-Package API
-===========
+{root_title}
+{root_title_underline}
 
 .. toctree::
    :maxdepth: 2
 
-   api/library_root
-   Full API <api/unabridged_api>
-   File structure <api/unabridged_orphan>
+   {package.name} <generated/index>
 
 Indices and Search
 ==================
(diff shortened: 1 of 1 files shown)


hidmic:
FYI @clalancette.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -78,23 +78,22 @@ def ensure_global(name, default):
...
         # Maps markdown files to the "md" lexer, and not the "markdown" lexer
         # Pygments registers "md" as a valid markdown lexer, and not "markdown"
         "lexerMapping": {{r".*\.(md|markdown)$": "md",}},
-        # This mapping will work when `exhale` supports `:doxygenpage:` directives
-        # Check https://github.com/svenevs/exhale/issues/111
-        # TODO(aprotyas): Uncomment the mapping below once the above issue is resolved.
-        # "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
-        #     lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
+        "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
+            lambda kind: [":content-only:", ":no-link:"] if kind == "page" else []),
```
aprotyas, line 96: IIRC, this also changes the behavior when there is a different (not `:doxygenpage:`) directive. As in, other directives such as `:doxygenfunction:` wouldn't render correctly with this mapping, and I think you have to provide mappings for each such directive. I will confirm whether that's true in a future comment.

aprotyas: :100: in general for the behavior this should give otherwise though.

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-10-28/22947/1

hidmic:
This is ready to go @clalancette.

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-11-18/23209/1

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -451,12 +443,17 @@ def generate_default_project_into_directory(self, directory):
             'package_authors': ', '.join(set(
                 [a.name for a in package.authors] + [m.name for m in package.maintainers]
             )),
-            'package_underline': '=' * len(package.name),
         }
 
         with open(os.path.join(directory, 'conf.py'), 'w+') as f:
             f.write(default_conf_py_template.format_map(template_variables))
 
+        root_title = f'Welcome to {package.name} documentation'
```
clalancette, line 451: I think this would read better as:

```
Welcome to the documentation for {package.name}
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -78,23 +78,22 @@ def ensure_global(name, default):
...
         "exhaleExecutesDoxygen": False,
         # Maps markdown files to the "md" lexer, and not the "markdown" lexer
         # Pygments registers "md" as a valid markdown lexer, and not "markdown"
         "lexerMapping": {{r".*\.(md|markdown)$": "md",}},
-        # This mapping will work when `exhale` supports `:doxygenpage:` directives
-        # Check https://github.com/svenevs/exhale/issues/111
-        # TODO(aprotyas): Uncomment the mapping below once the above issue is resolved.
-        # "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
-        #     lambda kind: [":project:", ":path:", ":content-only:"] if kind == "page" else []),
+        "customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
```
clalancette, line 95: Did we ever resolve @aprotyas issue with this?  Quoting:

```
IIRC, this also changes the behavior when there is a different (not :doxygenpage:) directive. As in, other directives such as :doxygenfunction: wouldn't render correctly with this mapping, and I think you have to provide mappings for each such directive. I will confirm whether that's true in a future comment.
```

clalancette (changes requested):
Now that we have the support in upstream exhale, we can move forward with this.  I have one suggestion inline

rosdoc2/verbs/build/builders/sphinx_builder.py
hidmic: It's funny, `exhale` documentation is misleading here. If `utils.makeCustomSpecificationsMapping` yields an empty list for the specs of given directive, it'll use Breathe defaults, which are not the same as Exhale defaults (see https://github.com/svenevs/exhale/blob/bbb5d71f126d1281a8c4f1729c218c236f44457b/exhale/utils.py#L185). I can change this to match Exhale defaults though.

rosdoc2/verbs/build/builders/sphinx_builder.py
hidmic: Agreed. See https://github.com/ros-infrastructure/rosdoc2/pull/33/commits/d94267302c5027101a95109df119f7bb365d37d3.

rosdoc2/verbs/build/builders/sphinx_builder.py
clalancette: It's not clear to me whether we should be going for the Breathe or Exhale defaults in that case.  @hidmic @aprotyas I would take whatever your recommendation here is.

aprotyas: The Exhale defaults are not bad, but those defaults are only for the `class` and `struct` kinds.
It would be nice to pay the one-time cost of a `rosdoc2_specsMapping` dictionary that lets us control the directive of all kinds that are of interest. It also makes for more extensibility:

```py
rosdoc2_specsMapping = {                                                                               
    'page': [':content-only:'],                  
    **dict.fromkeys(['class', 'struct'], [':members:', ':protected-members:', ':undoc-members:']),  
    # other kind-specs mapping                                                                         
}       

"customSpecificationsMapping": utils.makeCustomSpecificationsMapping(
    lambda kind: rosdoc2_specsMapping.get(kind, [])),
```

hidmic: That looks reasonable to me. See befbade175adfc6369f02be17718d570203f9499.

hidmic:
Rebased and ready for another round of reviews.

clalancette (approved):

aprotyas (approved):

PR #38, opened 2021-12-15
Fix flake8 errors throughout the tree.

clalancette:


rosdoc2/verbs/build/builder.py
```diff
@@ -47,27 +50,27 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
         May be overridden by downstream builders to selectively skip or rename files.
         """
         if os.path.exists(destination):
             raise RuntimeError(
                 f"Error integrating output from builder '{self.name} ({self.builder_type})': "
                 f"file '{common_suffix}' already exists in destination '{destination}'. "
-                f"This usually occurs when two builders generate the same file in "
-                f"the output directory.")
+                f'This usually occurs when two builders generate the same file in '
+                f'the output directory.')
```
hidmic, line 66: @clalancette nit: no need for these to be f-strings, regular strings would do. Same elsewhere for f-strings that lack `{...}` placeholders.

rosdoc2/verbs/build/impl.py
```diff
@@ -102,21 +103,20 @@ def prepare_arguments(parser):
 
 
 def main(options):
+    """Execute the program, catching errors."""
     try:
         return main_impl(options)
-    except Exception as e:
+    except Exception as e:  # noqa: B902
```
hidmic, line 109: @clalancette hmm, `B902`? The documentation I find suggests this is for methods (?).

test/test_flake8.py
```diff
@@ -26,7 +26,7 @@
 
 def test_flake8():
     style_guide = get_style_guide(
-        extend_ignore=['D100', 'D104'],
+        extend_ignore=['D100', 'D104', 'W605'],
```
hidmic, line 29: @clalancette why ignore `Ẃ605`?

clalancette: Ah, this one is tricky and I should have put a comment for it.

I think flake8 is misunderstanding the triple-quoted string at https://github.com/ros-infrastructure/rosdoc2/blob/b8f62374d74580662f617ff4ea01df5222252e81/rosdoc2/verbs/build/builders/sphinx_builder.py#L27-L131 , and thinking that there is a `\.` in a regex without an `r` at line https://github.com/ros-infrastructure/rosdoc2/blob/b8f62374d74580662f617ff4ea01df5222252e81/rosdoc2/verbs/build/builders/sphinx_builder.py#L104 .  So it was always giving me the [W605](https://www.flake8rules.com/rules/W605.html) warning.

That said, disabling it completely is indeed too big of a hammer.  I'll change it to just ignore W605 on that one line.

rosdoc2/verbs/build/impl.py
clalancette: As far as I can tell, `flake8-blind-except` throws a "B902 blind Exception: statement" error on a blind except like this one.  So this disables the warning just on this line.

rosdoc2/verbs/build/builder.py
clalancette: Good call, will fix.

clalancette:
See 98b8045, which should fix most of the issues pointed out above.

rosdoc2/verbs/build/impl.py
hidmic: Ha, [`flake8-bugbear`](https://pypi.org/project/flake8-bugbear/) overloads B902...

test/test_flake8.py
hidmic: @clalancette fyi, one alternative solution here would be to use raw triple-quoted strings `r"""..."""`.

hidmic (approved):

hidmic:
Pending green CI.

test/test_flake8.py
clalancette: Yeah, that's true.  I was a little hesitant to make changes that way, since there are 2 levels of replacement being done; one when we write out the string to a file from within `rosdoc2` (using f-strings), and then another at runtime when sphinx actually executes this file.  I *think* what you've suggested will work, but nothing is safer than a comment :).

clalancette:
We don't currently have CI setup here, but I did test this locally on a couple of packages and it seemed to have worked.  So I'm going to go ahead and merge this; thanks for the review!

PR #39, opened 2021-12-21
Use :content-only: as the doxygenpage directive.

clalancette:
According to the documentation, that is the syntax that is
supported.

The breathe documentation that talks about this is here: https://github.com/michaeljones/breathe/blob/master/documentation/source/directives.rst#doxygenpage


hidmic (approved):
:man_facepalming: oof, my bad. Thanks for the fix @clalancette !

PR #40, opened 2021-12-21
Add in a GitHub workflow.

clalancette:

Draft until I verify it actually works.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,9 @@ jobs:
         architecture: x64
 
     - name: Install pip dependencies
-      run: python -m pip install flake8 pytest
+      run: |
+       python -m pip install flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated flake8-docstrings flake8-import-order flake8-quotes pytest
+       python -m pip freeze
 
     - name: Run tests
       run: py.test --verbose test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,33 @@
+name: CI
+
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  unit_tests:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Setup python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: x64
+
+    - name: Install pip dependencies
+      run: |
+       python -m pip install flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated flake8-docstrings flake8-import-order flake8-quotes pytest
+       python -m pip freeze
```
hidmic, line 32: @clalancette maybe cache pip installed packages to speed things up a bit? See [actions/cache@v2](https://github.com/actions/cache/blob/main/examples.md#python---pip).

hidmic (commented):
Overall makes sense to me.

.github/workflows/test.yml
aprotyas, line 32: Just curious, what does this achieve?

Also, I know this is typically handled with `rosdep` but for this workflow's benefit, would a `requirements.txt` file containing the `pip freeze` for this project be a good idea?

clalancette: The idea with the `pip freeze` is really just to print out the versions of all of the pip packages are being installed.  It's useful if an error occurs and you are trying to recreate the environment locally.

`pip list` would do as well, but we tend to prefer `pip freeze` in our infrastructure for reasons I don't fully understand.

clalancette: Done in 7f8d743

clalancette: Switching to requirements.txt is a good idea.  I've done this in 7f8d743.

hidmic (approved):
LGTM

clalancette:
Thanks!  Merging.

PR #43, opened 2022-01-14
[Test] Ignore W503 warnings in Flake8

aprotyas:
Flake8 generates warnings for both W503 and W504 - line breaks before
and after binary operators respectively.
I came across both together when running CI for #28, and it's difficult
to satisfy both. Plus, it's documented that W503 is incompatible with
PEP 8. Reference: https://bugs.python.org/issue26763

As such, this PR chooses to ignore W503 warnings for this project.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+ignore = W503
+max-line-length = 100
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,0 @@
-[flake8]
-ignore = W503
-max-line-length = 100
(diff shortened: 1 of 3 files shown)


.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
```
clalancette, line 1: Instead of adding this new `.flake8` file, I'll suggest just adding this to the list of ignored flake8 things in https://github.com/ros-infrastructure/rosdoc2/blob/main/test/test_flake8.py

aprotyas: Good point. I've removed this file and ignored the warning in the test script instead (https://github.com/ros-infrastructure/rosdoc2/pull/43/commits/d60fb06db604962bb19ec54450baf3c777836d8e).

clalancette (approved):
Works for me, I don't mind disabling this one since upstream seems to be confused on what they want to do here.

PR #45, opened 2022-09-27
add default_config verb to create a default configuration file

ottojo:
Information about the config file format is hard to find, i think this makes it easier to use.

This adds a verb `default_config` which just writes the default config file already present in the code to a file, to be customized by the package author.


mikeferguson (approved):
I just tried this out (rebased on top of main) and I do think it would be a helpful addition to this tool (everything worked as advertised)

rosdoc2/verbs/default_config/__init__.py
```diff
@@ -0,0 +1,16 @@
+
```
Yadunund, line 1: Add copyright header

rosdoc2/verbs/default_config/impl.py
```diff
@@ -0,0 +1,31 @@
+import os
```
Yadunund, line 1: Add copyright header.

Yadunund (changes requested):
@ottojo thanks for the contribution and thank you @mikeferguson for reviewing this PR.

Left a couple of nitpicks to address before we can merge.

Alternatively @mikeferguson if you could open a PR with the `main` rebase and the copyright headers added, happy to merge that in as long as the origin commit authored by @ottojo is preserved.

ottojo:
Nice to see some activity here, I think package documentation is such an important aspect of ROS with much room for improvement!

I rebased the commit to latest main, and tested it myself as well.

rosdoc2/verbs/default_config/__init__.py
ottojo: I added a header, not sure if it is expected to dedicate copyright to OSRF, that can be done if desired.

tfoote (approved):
Thanks for pushing this forward

Yadunund:
Looks like our linters have flagged some formatting issues with the changes proposed. Do you mind addressing them?

ottojo:
should be fixed! @Yadunund

Yadunund (approved):

PR #47, opened 2023-01-27
Copy user files into docs_build so they can be referenced in conf.py

sloretz:
In support of https://github.com/locusrobotics/fuse/pull/278

`rosdoc2` `sphinx_builder` copies and wraps a user's `conf.py` in a separate `docs_build` directory outside the source tree. Unfortunately this breaks relative paths to things like images and `.rst` files. I don't see a way to get the correct absolute path because `exec(open("...").read())` is used by the wrapping `conf.py` to load the user's `conf.py`.

This PR solves it by copying the `user_sourcedir` content into the build directory.

An alternative solution might be to make the wrapping `conf.py` `import` the user's `conf.py` so that the user could make an absolute path from `__file__`.

@methylDragon FYI

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -487,6 +487,13 @@ class SphinxBuilder(Builder):
         intersphinx_mapping_extensions,
     ):
         """Generate the rosdoc2 sphinx project configuration files."""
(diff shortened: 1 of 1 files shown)


tfoote (approved):
This isn't optimal, but I don't have a cleaner suggestion.

clalancette:
Yeah, I'm going to close this as I think this is not needed anymore.

Issue #48, opened 2023-03-10
Checklist for supporting document generation for ament_python packages


Yadunund:
Automatic document generation for python (`ament_python`) packages in the ROS ecosystem has been a long pending ask and the purpose of this ticket is to serve as a checklist of PRs/patches to land to deliver this functionality.

### PRs/Code to merge into this repo in specific order:
- [x] https://github.com/ros-infrastructure/rosdoc2/pull/28: Bulk of the required code is in this feature branch
- [ ] https://github.com/rkent/rosdoc2/commit/3fb714120e594e80f58d8fe45658f369368fd13d: Port changes to `sphinx_builder.py` to append pkg source directory to system path when using autogenerated `conf.py`.
- [ ] https://github.com/ros-infrastructure/rosdoc2/pull/47
- [ ] TODO: Similar to #47, also copy `package_src_directory` made available from https://github.com/rkent/rosdoc2/commit/3fb714120e594e80f58d8fe45658f369368fd13d into `docs_build` such that  the python interpreter can access pakg modules when running the user written `conf.py` file
- [ ] https://github.com/Yadunund/rosdoc2/commit/af64741619e269897b90da1a1e5a3275d0cb71b6: Correctly populate `autodoc_mock_imports` with python modules that are listed as system packages in `package.xml` (eg. `python3-babeltrace`).
- [ ] https://github.com/Yadunund/rosdoc2/commit/226f90562be408979902d5102d68f96ced173ffc: Include `<doc_depend>` dependencies in list of modules to include in `autodoc_mock_imports`
- [ ] Bonus: Update sphix_builder.py to not always overwrite any user specific `autodoc_mock_imports` with `exec_depend` modules.

I've incorporated all these changes into temporary [branch](https://github.com/Yadunund/rosdoc2/tree/yadu/build-python-pkg) for my testing and I'm pleased to share that document generation has gone well for most of the `ament_python` packages in the `ros2` org. Along the way I've had to make some changes some packages to successfully generate documentation. Most of the changes are related to populating the `package.xml` with missing `exec_depends`. Without this, `autodoc` will try to import these dependency modules when importing the module to be documented. The downside of this is that cross-reference across pkgs is not yet supported but I'll discuss this and other todo's for the future later below. Other changes include fixes to docstrings to comply with sphinx/autodoc spec.

### PRs to merge to generate documentation without any warnings
- [x] https://github.com/ros2/demos/pull/606
[...]

PR #49, opened 2023-03-22
Additional changes to support ament_python pkgs

Yadunund:
This PR is a follow up to https://github.com/ros-infrastructure/rosdoc2/pull/28

It closes https://github.com/ros-infrastructure/rosdoc2/issues/48 by addressing all remaining issues (except the bonus item).

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -601,12 +601,19 @@ class SphinxBuilder(Builder):
         intersphinx_mapping_extensions,
     ):
         """Generate the rosdoc2 sphinx project configuration files."""
-        # Copy all user content, like images or documentation files, to the wrapping directory
+        # Copy all user content, like images or documentation files, and source files to the wrapping directory
         if user_sourcedir:
-            shutil.copytree(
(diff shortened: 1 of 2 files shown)


rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -27,6 +27,11 @@
 logger = logging.getLogger('rosdoc2')
 
 
+def esc_backslash(path):
+    """Escape backslashes to support Windows paths in strings."""
+    return path.replace('\\', '\\\\') if path else path
```
aprotyas, line 32: Couldn't instantiating `pathlib.Path` objects handle this?

tfoote: I like that idea but think that this is going into the config of the next stage where it is then needed to be escaped. 

I'd suggest considering switching to this as an enhancement for the future.

tfoote (approved):

rosdoc2/verbs/build/builders/sphinx_builder.py
Yadunund: I've opened a ticket to follow up on this https://github.com/ros-infrastructure/rosdoc2/issues/50

PR #51, opened 2023-04-02
Feature: Override build_type via rosdoc2 yaml

Yadunund:
This PR allows users to pass an `override_build_type` parameter via the rosdoc2.yaml settings.

I've found this to be the easiest way to get document generation for `rclpy` like packages to work. ie, where the `build_type` in `package.xml` is `ament_cmake` but we want autodoc generation of the python modules within `rclpy/rclpy`.


tfoote (approved):
This looks reasonable as an option to add.

clalancette (approved):

Issue #42, opened 2022-01-06
Process user's conf.py and index.rst through templates like the default version


rkent:
I've been experimenting with rosdoc2 for the last month, because I want to make sure that I do proper documentation as I convert my robot to ros2. I have a lot of suggestions, but really the most import is to use templates with user content.

Currently, rosdoc2 can create a default conf.py and index.rst, and when it does so it uses python's format_map method with some template variables. Currently user-provided conf.py and index.rst do not have the same ability to use the template variables.

As a user, there are a few reasons why I want to have the same ability to use the template variables. The two main reasons are:
* I want to keep my documentation in-sync with the various things like package.version and package.description
* When I customize conf.py or index.rst, I want to start with a standard file, and possibly do minor changes to it.

In addition, if rosdoc2 is expanded to include more documentation beyond the current doxygen material, there are a lot more items that could be available in the template variables. In my demo, I've included:
* Links to the standard package documentation like README and LICENSE
* A link to index.ros.org, if it exists
* Links to soon-to-be-automatically-generated message format descriptions
* Links to documentation included in standard /doc folder

In PR #28 the template was expanded to include optional Python documentation. The same issue will be faced when msg, service, and action descriptions are added, so the available template variables are going to get more extensive.

The actual templating library is not important. I am not a fan of using python's format_map because it forces me to to do changes to python code away from the template variables to escape common constructs like {}. I should not have to mess with my code outside of adding the template variables in order to use the templating system. In my demo, I used jinja. But which one to use is not important.

My current demo is at https://github.com/rkent/rosdoc2/tree/index_links The default index.rst is included as a file rather than embedded in the code as [index.j2](https://github.com/rkent/rosdoc2/blob/index_links/rosdoc2/verbs/build/builders/index.j2)

My demo package documentation can be seen [here](https://rosdabbler.github.io/fqdemo/fqdemo/). It's a WIP though.

If this is an acceptable direction, I can submit some pull requests to implement.

rkent:
This is going to be a pivotal issue for me to make progress with rosdoc2. I'll be doing a PR to suggest adding this feature, but let me show you where I want to go.

[...]

PR #60, opened 2024-02-16
add more default configuration to sphinx conf.py

ottojo:
Add values extracted from package.xml to the wrapping sphinx config.py. This eliminates the need for packages with custom config.py to duplicate author, copyright and version information from package.xml.

A more comprehensive proposal was made in #42, but this part (keeping package.xml and config.py in sync) was what i came across today, and what is solvable a lot easier...


rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -72,6 +73,10 @@ def ensure_global(name, default):
 
 ensure_global('rosdoc2_settings', {{}})
 ensure_global('extensions', [])
+ensure_global('project', "{package_name}")
+ensure_global('copyright', "{year}, {package_licenses}")
```
tfoote, line 77: The copyright is usually the year and the copyright holder(s), not the year and the types of licenses. Unfortunately this isn't something that is captured in the package.xml there can be one or many copyright holders. I'd recommend skipping this one instead of doing a partial job. Also it's not clear to me if the current year is not technically right for this if there's no edits in the current year.

tfoote (changes requested):
Thanks extending it this way will help get more visibility. I'd like to pull back the copyright info though as omitting it is better than giving partial info.

rosdoc2/verbs/build/builders/sphinx_builder.py
ottojo: makes sense to me, i removed it for now

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -643,7 +647,12 @@ def generate_wrapping_rosdoc2_sphinx_project_into_directory(
             'user_conf_py_filename': esc_backslash(
                 os.path.abspath(os.path.join(user_sourcedir, 'conf.py'))),
             'breathe_projects': ',\n'.join(breathe_projects) + '\n    ',
-            'intersphinx_mapping_extensions': ',\n        '.join(intersphinx_mapping_extensions)
+            'intersphinx_mapping_extensions': ',\n        '.join(intersphinx_mapping_extensions),
+            'package': package,
+            'package_authors': ', '.join(set(
```
tfoote, line 652:
```suggestion
            'package_authors': ', '.join(sorted(set(
```

One tweak to sort the output for repeatability.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -643,7 +647,12 @@ def generate_wrapping_rosdoc2_sphinx_project_into_directory(
...
+                [a.name for a in package.authors] + [m.name for m in package.maintainers]
+            )),
```
tfoote, line 654:
```suggestion
            ))),
```

Close sort parens.

tfoote (commented):
Thanks for the revision. One more tweak to make the output repeatable and more readable.

tfoote (approved):

PR #61, opened 2024-02-17
include package readme in landing page

ottojo:
I implemented this for a specific package in https://github.com/teamspatzenhirn/rviz_2d_overlay_plugins/pull/19, and saw that a feature like this was already requested/investigated in #22 . This does work with relative paths to images, i have not tested it for links to more markdown pages.

This is just a rough hack so far, at least the path-handling has to be cleared up a bit i think :smile: 

Any feedback and discussion is appreciated, i think including the readme is a good "random try" at pulling in existing information into the docs for packages without any further configuration, and might be desired to reduce duplication even once a package adds more docs (although i think lowering the barrier to and establishing a culture of deliberately writing docs beyond the readme should be the main goal).

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -322,6 +322,13 @@ index_rst_template = """\
 {root_title}
 {root_title_underline}
 
+{package.plaintext_description}
+
+{package_readme_include}
+
+API Docs
+========
+
 .. toctree::
    :maxdepth: 2
 
@@ -332,6 +339,22 @@ Indices and Search
 
 * :ref:`genindex`
 * :ref:`search`
+
+"""
+
+readme_include_md_template = """\
+```{{include}} {readme_file_name}
+:relative-images:
+```
+
+"""
+
+readme_include_rst_fragment = """\
+Package README
+==============
+
+.. include:: readme_include.md
+   :parser: myst_parser.sphinx_
 
 """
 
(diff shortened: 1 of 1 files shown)


rkent:
I did something very similar in my (abandoned) work two years ago to try to improve rosdoc2. The initial commit to process standard documents is https://github.com/rkent/rosdoc2/commit/4b2872b38749eb48a111f9b4233610129a14ae28  The last work on this branch is https://github.com/rkent/rosdoc2/commits/rkent/jinja-templates

Some general comments here. There are not just markdown READMEs, but also rst versions. It is fairly trivial so support both. See for example https://github.com/ros-drivers/nmea_msgs but there are many others.

Also, README is not the only standard document. While you are searching for that, it is fairly trivial to generalize it to also get other standard documents: ["readme", "license", "contributing", "changelog"]. Some of those might end up being plain text, so it is good to support that as well.

rkent:
This should be closed as completed in PR #98

ottojo:
Nice, great to see this idea implemented properly! It's also great to see a lot of activity around rosdoc2 lately. Is there somewhere to read up on the long term goals or current plans/roadmap?

rkent:
I could not really concretely discuss plans until I knew if the concept of jinja templates was going to fly. Well it flew, so now the big picture is much clearer. I'm also almost done with the proposed changes, so at this point it will be better to spend my time documenting what is implemented.

This work has been landing stuff that I did two years ago, and it is mostly landed. You could see what the plan was two years ago [here](https://rosdabbler.github.io/fqdemo/fqdemo/). We're pretty close to that now. That repo was my attempt to understand the "proper" way to structure a ROS2 project back when I was trying to convert my hobby robot from ROS1 to ROS2. rosdoc2 was clearly not in shape, so I started working on it.

PR #73, opened 2024-03-22
Allow running as module

rkent:
When developing rosdoc2, it can be convenient to run it as a module directly rather than having to install each time, or trying to make an editable install work. This minor change should allow it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,12 @@ rosdoc2 open ./doc_output/index.html
 
 For more advanced usage see the documentation.
 
+It may be helpful during rosdoc2 development to run a version of rosdoc2 without installing it. This can be done
+(after doing an initial normal install to make sure prerequisites are available) by running, from the rosdoc2 main directory:
+```bash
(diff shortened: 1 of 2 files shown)


tfoote (commented):
Does this have significant benefits over using `--editable` from pip install?

rkent:
Maybe I am doing wrong, but whenever I use --editable it demands that I be using a venv, which I don't typically do. That adds extra overhead to running things.

Other than that, the main advantage is that you don't have to modify your rosdoc2 install.

It's not a big deal, really with the testing infrastructure I am proposing the vast majority of development takes place there. But if I am trying to fix an issue that exists in a package, and I don't want to add that issue to a test case, this could be useful.

tfoote (approved):
I guess I always do venv based development these days as I jump environments all the time and often need to have whole chains of dependencies different. If this enables working on this without it though. The change is small enough in unobtrusive enough that it seems fine to me to support that approach too.

PR #74, opened 2024-03-25
Try out container matrix

tfoote:


rkent:
Cool, I didn't know ```container: ['ros:iron', 'ros:rolling']``` was an option

tfoote:
I wasn't sure either. You can do arbitrary keys. That's why I tried it out in the PR before suggesting it in #72 I think that the setup/python might not be necessary with this new base image too.

tfoote:
As this was a test, I'll close this and we can integrate the work into the iteration in #72

Issue #54, opened 2023-04-29
message only packages are failing to generate documentation


tfoote:
A job like Rdoc_common_interfaces https://build.ros2.org/view/Rdoc/job/Rdoc__common_interfaces__ubuntu_jammy_amd64/

It is building sensor_msgs because they have an include folder and so doxygen is triggered. But all the other packages in that repository generate no content.

Example: https://build.ros2.org/view/Rdoc/job/Rdoc__common_interfaces__ubuntu_jammy_amd64/26/console

rkent:
Would it be acceptable if these packages generated documentation for the msgs that was simply a copy of the msg file, with no parsing? That's how I did it in my earlier iteration.

rkent:
Hmmm, no response, so I guess I'll just do this in a PR.

I could never figure out a non-tacky way of generating cross references to msg and srv files, which would be useful. But just showing the definitions as text files would be better than nothing.

Yadunund:
> Would it be acceptable if these packages generated documentation for the msgs that was simply a copy of the msg file, with no parsing? That's how I did it in my earlier iteration.

I think this is a good start.


In the long run, it would be great to also generate documentation for code generated by the various `rosidl_generators` but this is non-trivial since we do not build the packages before running `rosdoc2`.

rkent:
My experience is that just showing the text works pretty well. You can see an example at https://rosdabbler.github.io/fqdemo/fqdemo_msgs/generated/msg/NumPwrData.html from my experimental version of rosdoc2.

The tricky part is trying to get cross references to msg classes to work. That experiment does it by declaring the message to be a python class that can be cross-referenced, but I'm not sure that is a good solution.

Anyway I'll propose a patch here once some of my other PRs clear.

PR #76, opened 2024-03-27
Support msgs, services, and actions

rkent:
This will show messages, services, and actions (with just their raw text) in packages that already have C++ content. Unfortunately it does not resolve issue #54, that will take additional effort.

In the test cases, full_package has messages, services, and actions that can show how this works.

The majority of real packages that have msg definitions seem to be in msg-only packages, which will still fail. So this PR does not really solve the real-world problem yet, but I will submit a PR to address msg-only packages. It does not really matter to me if this is reviewed and landed first, or if you wait until I resolve the case of message-only packages, it is the same amount of work for me. Your choice.

### DIFF
--- a/rosdoc2/verbs/build/build_context.py
+++ b/rosdoc2/verbs/build/build_context.py
@@ -30,3 +30,4 @@ class BuildContext:
         self.python_source = None
         self.always_run_doxygen = False
         self.always_run_sphinx_apidoc = False
+        self.ament_cmake_python = False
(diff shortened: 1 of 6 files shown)


tfoote (approved):
Thanks I'm here and reviewing so might as well bring this in with the smaller incremental change. This is great to have these rendering with more than just the raw github link like we currently do in index.ros.org

ros-discourse:

ros-discourse:

https://discourse.ros.org/t/plans-for-ros-2-documentation-particularly-api-docs/28638/23

PR #78, opened 2024-03-27
Support only messages

### LINKED ISSUE #54: message only packages are failing to generate documentation

rkent:
This was less complex than I feared. Test case ```only_messages``` shows how this works.

I find the whole process of deciding what to run to be hard to understand. Here I change the decision to run breathe based on the existence of the doxygen output, while previously it was done based on build type. Both seem a little too much like black magic to me. I feel like it could use a refactoring, but maybe the complexity is just inherent and hard to improve.

Fixes #54


tfoote (approved):
Using the logic to run breathe if doxygen was run makes sense. It's specifically for integrating the doxygen content into a sphinx runner. The previous logic was a proxy for where we were expecting doxygen to be run. (Aka c++ build types) But directly detecting that there's content available makes sense and is likely more robust.

I've tested that this is working for packages known to be previously failing.

tfoote:
Confirmed building on the bulidfarm with message only packages: https://docs.ros.org/en/rolling/p/trajectory_msgs/

Yadunund:
🎉 🎉 🎉

ros-discourse:

ros-discourse:

https://discourse.ros.org/t/plans-for-ros-2-documentation-particularly-api-docs/28638/23

PR #84, opened 2024-03-29
Deprecate --build-directory option (resolves #79)

### LINKED ISSUE #79: Remove unused --build-directory option
AFAICT, the --build-directory option is unused. If someone can figure out why it is there, please tell me. If not, we should remove it.

The correct option for the "directory where document builds are done" is --doc-build-directory. I wish it was "--build-directory" but I think that ship has sailed, we should leave it a "--doc-build-directory"

rkent:
I didn't set a specific date, just "future".

### DIFF
--- a/rosdoc2/verbs/build/impl.py
+++ b/rosdoc2/verbs/build/impl.py
@@ -61,7 +61,7 @@ def prepare_arguments(parser):
     parser.add_argument(
         '--build-directory',
         '-b',
-        help='UNUSED, to be removed in the future',
+        help='UNUSED, to be removed at some time after September 1st, 2024',
     )
     parser.add_argument(
         '--install-directory',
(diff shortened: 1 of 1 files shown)


rosdoc2/verbs/build/impl.py
```diff
@@ -62,7 +61,7 @@ def prepare_arguments(parser):
     parser.add_argument(
         '--build-directory',
         '-b',
-        help='build directory of the package',
+        help='UNUSED, to be removed in the future',
```
tfoote, line 64:
```suggestion
        help='UNUSED, to be removed at some time after September 1st, 2024',
```

tfoote (commented):
We don't have a formal deprecation cycle so I picked a date to give people warning but also make sure it doesn't have to linger too long. And hopefully this will be a reminder to come back to it.

PR #85, opened 2024-03-29
Restrict interfaces by directory (fixes #82)

### LINKED ISSUE #82: message generation is picking up and documenting messages that are supposed to be hidden
Packages like rclcpp have test content in sub folders which are not advertised messages. However they're being added to the package documentation. 

image

```
rclcpp$ grep -rI LargeMessage 
test/rclcpp/test_publisher_with_type_adapter.cpp:struct TypeAdapter<std::string, rclcpp::msg::LargeMessage>
test/rclcpp/test_publisher_with_type_adapter.cpp:  using ros_message_type = rclcpp::msg::LargeMessage;
test/rclcpp/test_publisher_with_type_adapter.cpp:  using StringTypeAdapter = rclcpp::TypeAdapter<std::string, rclcpp::msg::LargeMessage>;
test/rclcpp/test_publisher_with_type_adapter.cpp:  using StringTypeAdapter = rclcpp::TypeAdapter<std::string, rclcpp::msg::LargeMessage>;
test/rclcpp/CMakeLists.txt:  ../msg/LargeMessage.msg
```

We should limit the discovery of messages to be in the msg directory.

rkent:
I'm willing to restrict the search to certain directories, but it would be good if there was a reference that supports that. Otherwise this seems to work.

### DIFF
--- a/rosdoc2/verbs/build/generate_interface_docs.py
+++ b/rosdoc2/verbs/build/generate_interface_docs.py
@@ -44,8 +44,12 @@ toc_fm_rst = """\
 
 def _find_files_with_extension(path, ext):
     # Partly adapted from https://github.com/ros-infrastructure/rosdoc_lite
+    # and https://stackoverflow.com/questions/19859840/excluding-directories-in-os-walk
     matches = []
-    for root, _, filenames in os.walk(path):
+    excludes = ['test', 'doc']
+    for root, dirs, filenames in os.walk(path, topdown=True):
+        # exclude directories
+        dirs[:] = [d for d in dirs if d not in excludes]
         for filename in fnmatch.filter(filenames, f'*.{ext}'):
             matches.append((os.path.splitext(filename)[0], os.path.join(root, filename)))
     return matches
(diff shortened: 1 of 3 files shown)


rkent:
Sorry, I did not see you comment on issue #82 . I suppose you want it restricted to only look at the expected directory, starting from the package directory?

tfoote (changes requested):
Yes please do restrict it to the msg, srv, and action subdirectories. 

I improved the docs to make that explicit https://github.com/ros2/ros2_documentation/pull/4246



And there's already tools for walking and finding messages services and actions in python that we could reuse if it was already built as the messages would be registered with the index.

I think you can use get_*_interfaces from rosidl_runtime_py https://github.com/ros2/rosidl_runtime_py/blob/df2361f64ecb85f65f4bf22e804424e0f1ac4197/rosidl_runtime_py/get_interfaces.py#L72

In the same way that `ros2 interface package std_msgs` works. https://github.com/ros2/ros2cli/blob/ce945cd70d83a9ff41f08594e40ea96a0e6ac308/ros2interface/ros2interface/verb/package.py#L20

But for now just walking makes sense.

tfoote:
Closing as superseded by #87

PR #86, opened 2024-03-29
Don't tell users to build first (resolves #80)

### LINKED ISSUE #80: Should packages be built locally before documenting?
The current rosdoc32 instructions say:
```
 To generate the documentation for almost any ROS 2 package, first build the package locally, ...
```
Is it necessary or useful to build the document locally first? Does the build farm do that? If not, we should change the instructions.

rkent:
Just remove the text.


tfoote (approved):
Thanks for the cleanup

PR #87, opened 2024-03-29
Specify interface directories

### LINKED ISSUE #82: message generation is picking up and documenting messages that are supposed to be hidden

rkent:
This also fixes #82, and is an alternative to PR #85.

Note the service definition that was formerly in test case full_package/full_package/srv (the python directory) is now in full_package/srv, and will fail if left in the previous position.


tfoote (approved):
Thanks for making this more explicit and adding the test for exclusions.

PR #89, opened 2024-04-02
Limit navigation_depth to 4, fixes #56

### LINKED ISSUE #56: Generated content uses a lot of disk space for a small subset of packages.
We got an alert on disk space from our hosts. And a quick look at our documentation usage showed most packages well under 100MB. But there were quite a few with generated content much bigger. 

Here's the top offenders in rolling

```
99M smacc2
100M rmf_traffic
102M rosidl_runtime_c
105M osrf_testing_tools_cpp
106M sm_dance_bot_warehouse_2
132M nav2z_client
133M eigenpy
135M sm_advanced_recovery_1
145M rviz_default_plugins
160M rmw
167M rcl
169M rmf_utils
319M hpp-fcl
869M rclcpp
1.2G sm_multi_stage_1
1.3G sm_pack_ml
1.8G proxsuite
2.0G vitis_common
7.1G fastrtps
22G total
```

It would be good to understand why these are blowing up and keep that from happening.

rkent:
Fixes #56 

The html_theme_option 'navigation_depth' has a large impact on the size of the generated output. This patch changes that from infinite (-1) to 4.

Size effects for rmf_utils iron:
depth = 2: 60M
depth = 3: 61M
depth = 4: 112M
depth = -1: 332M

With depth = 4, the C++ API section in the toc shows like:
C++ API
--Full C++ API
----Classes and Structs
------Class LazyExpression

That is, the toc shows the class name, but not any details of the class. You need to click on the name to reach the class page to get that.

I think that is the right amount of detail, plus it is reducing the total file space to 1/3 of the current value.

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -278,7 +278,7 @@ html_theme_options = {{
     # Toc options
     'collapse_navigation': False,
     'sticky_navigation': True,
-    'navigation_depth': -1,
+    'navigation_depth': 4,
     'includehidden': True,
     'titles_only': False,
 }}


tfoote (approved):
This is great to bring the disk space usage way down and also makes the navigation a little less cluttered. I see the collapse_navigation is False though the navigation seems to collapse by default now. But everything is still completely navigable. local testing on rclcpp took it from 1.6GB to 444MB on my machine.

tfoote:
Also side note it cuts the build time down to close to a third of the time as well for rclcpp. From 31 to 12 minutes
image

PR #90, opened 2024-04-03
Include user documents

### LINKED ISSUE #83: Show user provided documentation from /doc (and maybe from /docs)
I've been testing my implementation of showing user docs. This is likely to be one of the more significant (and potentially controversial) implemented features.

Ultimately I think this is going to proceed by doing a trial implementation, running on an existing distro (just limited to packages that have relevant user documentation), and allowing us to examine the effects. But that is not a trivial project, so I want to see if there are any opinions here first.

Some issues:
- package.xml supports the doc_depend tag. Is that used anywhere, if so how? If that is for package docs, how does that interact with rosdoc2?
- [Developer guidelines](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html) state that the /doc directory should contain all the documentation. But some packages use /docs instead. Do we support /doc and /docs?
- Should we attempt to render user docs that we find, even if they were intended to be rendered with another system?

rkent:
fixes #83

This PR is fairly significant, because it adds user documentation from the /doc folder into rosdoc2.

Notes:
- The PR includes some previously submitted PRs (like PR #88) so I might need to rebase it before final.
- Assuming PR #88, the items generated into the build document only show up in the docs_build folder, so there is no good reason to tag those with a 'generated/' directory. Also, as we add more items to rosdoc2, there are going to be additional things added. So instead I am using named directories for added things. Here I use 'user_docs', I also renamed the interface landing directory to 'interfaces'
- The style that I have been using for the parent index has been 'category/details', where category is 'C++ API' or 'Python API' for example. Here I continued that, using 'Documents' as the category. Virtually everyone who customizes their index.rst puts each document instead at the top level of the index. But they are controlling the details of that. If there are a lot of documents, other important categories would get lost, so I think using a containing index entry "Documents" is preferable when that is going to be an unordered list of documents from a folder.
- What to do with documents in subdirectories? Items like /doc/subdirectory/thedocument.rst are placed in the index with the name 'subdirectory/' with a corresponding .rst document created to list all of those documents.

To help with review, I processed packages that have /doc folders, and uploaded the results to http://test1.rosdabbler.com  You can also get some idea looking at the rendering of full_package in the test suite.

Much of the time the result is favorable. However there are a number of issues that can be seen.
- Internal anchors are not working (like ```## Overview [⬆](#swri_dbw_interface)```
- Some documents include embedded math or latex that is not being rendered properly, and instead shows up as embedded code gobbledygook, (like ```$$\begin{equation}\label{eq:QP}\tag{QP} ```)
- Some documents include cross-reference to other documents that assumes things that don't exist in our implementation.

Some of these issues might be fixed with adding additional options to myst, or with other fixes. I am going to recommend though that we deal with those in followup PRs, rather than try here to fix issues in specific package renderings.


tfoote:
The others are merged and this now has a conflict, probably the fixups from before the merge. 

The example output looks great such as rclcpp's API review just popping out http://test1.rosdabbler.com/rclcpp/  

One I noticed as the fuse_doc directory here: http://test1.rosdabbler.com/fuse_doc/ that appears to have overridden the wrapper. Is that from a different test or did it's sphinx take over the wrapped invocation?

rkent:
fuse_doc has completely replaced our generated index.rst and conf.py with their own. See http://docs.ros.org/en/iron/p/fuse_doc/ it should be no different.

This should be ready for review again.

tfoote (approved):
Thanks for pushing on this. The updates to bring in these additional resources is great.

PR #93, opened 2024-04-09
Fix flow-map syntax of rosdoc2.yaml

christophfroehlich:
There are some issues with the yaml syntax of the default config.

I created a default rosdoc2.yaml with `rosdoc2 default_config --package-path src/realtime_tools/`, but the linter complains

```
Flow map in block collection must be sufficiently indented and end with a }YAML
Unexpected flow-map-end token in YAML stream: "}"YAML
```

Increasing the indent fixes this for me.

### DIFF
--- a/rosdoc2/verbs/build/inspect_package_for_settings.py
+++ b/rosdoc2/verbs/build/inspect_package_for_settings.py
@@ -80,13 +80,13 @@ builders:
     - doxygen: {{
         name: '{package_name} Public C/C++ API',
         output_dir: 'generated/doxygen'
-    }}
+      }}
     - sphinx: {{
         name: '{package_name}',
         ## This path is relative to output staging.
         doxygen_xml_directory: 'generated/doxygen/xml',
         output_dir: ''
-    }}
+      }}
 """
 
 


rkent:
I don't understand why the test failed, it passes locally for me. Looks to me like some sort of unrelated one-time system glitch.

I did not see any specification of yaml style in the [ros code style specification](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html). Is there a ros standard yaml linter? I am not aware of any. What linter did you use?

The proposed change seems harmless, and does conform to most examples I have seen of yaml, but I am by no means an expert on yaml code style or syntax.

christophfroehlich:
I refer to the [redhat.vscode-yaml](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) extension. I don't know how strict the specification is, or if different linters can have different interpretations.

tfoote:
CI was broken due to the noble upgrade under the hood. It was fixed in #96 can you rebase? I'm not sure if we have a formal linter formatting. I think we typically use the default output from the python library.

tfoote (approved):
Thanks for iterating and cleaning this up.

PR #95, opened 2024-04-10
Use venv to run tests in github workflow

rkent:
Here's my attempt at fixing the pip failure.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,9 @@ on:
 jobs:
   unit_tests:
     name: Unit Tests
+    defaults:
+      run:
+        shell: bash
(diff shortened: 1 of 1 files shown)


tfoote:
If we're going to use a venv we should use the one natively from github actions instead of hard coding it ourselves: 

I got it working here: 
https://github.com/ros-infrastructure/rosdoc2/pull/96/commits/e4ab62a55b5d04444e3303d6173a321b972cb3b1 in my tirals in #96 

However I was really hoping to keep using the system version as that's what the users will experience unless we change the instructions to explicitly use a venv.

rkent:
OK we're not going to do it this way, so I will close this.

PR #96, opened 2024-04-10
Trying to resolve noble based CI fixes

tfoote:


rkent:
You would probably know better than me, but after reading the material on the changes to python, it seems to me the instructions to users needs to be in ubuntu 24 (and no reason not to do that earlier):
- install rosdoc2 using pipx
- test rosdoc2 with a venv and pip install -e .[test]

I really don't think that ```PIP_BREAK_SYSTEM_PACKAGES: 1``` is what we are going to recommend to users. Is there any reason to prefer that over venv in the github runner?

tfoote:
Yeah we don't want to ask users to do that. It's only necessary for flake8 unfortunately. 

There's a new proposed approach at: https://github.com/ros-infrastructure/rosdep/pull/930 I haven't had a chance to review yet.

tfoote:
I'm going to squash this as it is to get is back to green. We're pushing hard to get the noble image to have flake8 in a working format via apt before Noble is finalized.

ros-discourse:

https://discourse.ros.org/t/plans-for-ros-2-documentation-particularly-api-docs/28638/32

ros-discourse:

https://discourse.ros.org/t/plans-for-ros-2-documentation-particularly-api-docs/28638/34

PR #97, opened 2024-04-11
Use venv for tests, pipx for install

rkent:
This approach to fixing the issues in Ubuntu 24 is to use pipx for a normal installation of rosdoc2, but tests using venv.

Using pipx required some changes to sphinx_builder. The venv that is setup behind the scenes for pipx did not translate into a python subprocess, so runs of sphinx-build failed because of a missing exhale dependency. The solution to that was to follow the same process I do in tests, namely to run sphinx directly as a python module and not as a subprocess. That did not seem to be needed for sphinx-apidoc.

pipx has enough issues that it was worth adding directly to the test suite, simulating what we expect for  user install.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,9 @@ on:
 jobs:
   unit_tests:
     name: Unit Tests
+    defaults:
+      run:
+        shell: bash
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


rkent:
Just a note. This patch is more of a demonstration of an alternate approach than a serious request to land code. The larger ROS2 project will need to decide if ```PIP_BREAK_SYSTEM_PACKAGES: 1``` is the recommendation to users, or if the use of venv and pipx, as recommended in pip's ["Externally Managed Environments"](https://packaging.python.org/en/latest/specifications/externally-managed-environments/#externally-managed-environments) is to be followed.

What this patch demonstrates is that rosdoc2 can, with minor changes, be used with pipx. We should probably just do those changes anyway regardless of the overall project decision.

tfoote:
We should probably converge our systems to using pipx in the future. For the moment I'd suggest we hold off until we're ready to do it more generally.

rkent:
I'm going to close this, but rework it to only include the changes (and tests) for pipx compatibility. I'll explain why in a new PR.

Issue #106, opened 2024-04-17
Broken with Sphinx 7.3


tfoote:
Between the CI of #99 and test build on merge the CI appears to have broken

Passing:

https://github.com/ros-infrastructure/rosdoc2/actions/runs/8699265272/job/23857580345

`Sphinx==7.2.6`

Broken: 

https://github.com/ros-infrastructure/rosdoc2/actions/runs/8726584290/job/23942199944

`Sphinx==7.3.5`

7.3 has come out in the last 2 days with several patch releases:

https://pypi.org/project/Sphinx/#history

![image](https://github.com/ros-infrastructure/rosdoc2/assets/447804/9b6d69ce-71a3-464a-bc70-0d52e4ec5a11)

The core error:
```
Exception occurred:
  File "/usr/local/lib/python3.10/dist-packages/breathe/directives/function.py", line 161, in DoxygenFunctionDirective
    def _parse_args(self, function_description: str) -> Optional[cpp.ASTParametersQualifiers]:
AttributeError: module 'sphinx.domains.cpp' has no attribute 'ASTParametersQualifiers'
The full traceback has been saved in /tmp/sphinx-err-3wec0qsg.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```

tfoote:
There's a ticket upstream with a comment saying that Breath is broken too: https://github.com/sphinx-doc/sphinx/issues/12295#issuecomment-2061206831

tfoote:
Looks like this is patched in main as of an hour ago: https://github.com/sphinx-doc/sphinx/pull/12298 hopefully this will come out shortly. And we can wait a bit for the fix.

rkent:
Working now with Sphinx 7.3.6 released.

tfoote:
Great, I'm retriggering https://github.com/ros-infrastructure/rosdoc2/actions/runs/8726584290 to validate. And it passed!

PR #107, opened 2024-04-17
Try out using new apt-based installation

tfoote:
Flake8 landed on noble. Maybe this won't work on iron though? Opening this to see how CI looks.

But it's currently broken due to #106 

Working to roll back the changes necessary in #96

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,14 +53,24 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Install pip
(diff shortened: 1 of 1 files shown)


tfoote:
@rkent With Noble landing we can simplify the install instructions again.

rkent:
@tfoote this works with flake, but we still have to face the issues with sphinx addons like breathe and exhale, right?

tfoote:
Yes, they will still come into the pip environment.

Issue #110, opened 2024-04-18
Temporary paths are leaking into generated output /tmp/ws/src/<PACKAGE>/...


tfoote:
As an example: https://docs.ros.org/en/rolling/p/fastrtps/generated/file__tmp_ws_src_fastrtps_include_fastdds_rtps_security_accesscontrol_EndpointSecurityAttributes.h.html#definition-tmp-ws-src-fastrtps-include-fastdds-rtps-security-accesscontrol-endpointsecurityattributes-h

`/tmp/ws/src/fastrtps` appears in many places in the docs. Hopefully we can tweak a setting to truncate this in the generation stage. 

![image](https://github.com/ros-infrastructure/rosdoc2/assets/447804/174c4de8-6afe-4b0b-958d-53f76da3de62)

tfoote:
Duplicate of #71 fixed in #111

PR #111, opened 2024-04-18
Set STRIP_FROM_PATH in Doxyfile

### LINKED ISSUE #71: Directories reports an absolute directory rather than relative, and in worst case shows all parent directories
When you run rosdoc2 on a C++ project, the Directories html shows the full path to the original directory, which is typically in some sort of staging location. It should show the relative directory.

For example, In https://docs.ros.org/en/rolling/p/rosbag2_cpp/generated/dir__tmp_ws_src.html we have:
```
Directory path: /tmp/ws/src
```
which is exposing the buildfarm directories used here. '/tmp/ws' is of no interest, it should just say 'src'.

Worse, if the doc build directory is specified, then the Directories list will go all the way back to the drive root. So if I run rosdoc2 in /home/kent/github/ros2/demos/demo_nodes_cpp with:
```
rosdoc2 build -p . -d /tmp/rkent/docs_output
```
Then the directories listing at ```http://localhost:8001/docs_output/fqdemo_nodes/generated/index.html#directories``` shows:
```
Directories
    [Directory home]
    [Directory kent]
    [Directory github]
    [Directory ros2]
    [Directory demos]
    [Directory demo_nodes_cpp]
    [Directory include]
    [Directory demo_nodes_cpp]
```
Instead, it should just show ```include``` and ```demo_nodes_cpp```
The fix to this is relatively straightforward: set STRIP_FROM_PATH in the Doxyfile. I'll give a PR with that change, but I would like to add a test for it so I'll wait for the [my previous PR.](https://github.com/ros-infrastructure/rosdoc2/pull/70)

rkent:
Fixes #71 Isn't this also the same issue as issue #110?

### DIFF
--- a/rosdoc2/verbs/build/builders/doxygen_builder.py
+++ b/rosdoc2/verbs/build/builders/doxygen_builder.py
@@ -38,6 +38,7 @@ GENERATE_LATEX         = NO
 
 MACRO_EXPANSION        = YES
 EXPAND_ONLY_PREDEF     = YES
+STRIP_FROM_PATH        = {package_directory}
 """
 
 EXTENDED_DOXYFILE = """\
--- a/test/test_builder.py
+++ b/test/test_builder.py
@@ -85,7 +85,7 @@ def do_test_package(
     file_excludes=[],
     links_exist=[],
     fragments=[],
-) -> None:
+) -> htmlParser:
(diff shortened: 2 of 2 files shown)


tfoote (approved):
Yes, sorry that is a duplicate issue. This is great to resolve.

PR #112, opened 2024-04-18
Add package description to home page

rkent:
Note I used package.description rather than package_description. I intend to refactor the template variables to simplify them, in anticipation of documenting them for package authors to use if modifying .jinja files. There is no need to include package attributes that are available directly from the package.


tfoote (approved):
This is great to add.

PR #113, opened 2024-04-18
Add url links in home page

rkent:
This adds the urls from the package to the home page.

Note these are not included in the index in this version. I could do that, but I think I would need to first generate an rst page containing just these links.

In any case, the home page is getting cluttered now with lots of stuff that is also accessible from the index. I plan to do a revamp of that, hiding many of the items that are in the index from the home page, and simplifying the home page. In that case these URLs would only be on the home page, and not in the index. I think that is the right decision, but it is a matter of opinion.


tfoote (approved):
+1 for adding this in. We are getting a bunch of info. It might make sense to have a standard info box that collects things like package.xml content and renders them consistently.

PR #115, opened 2024-04-18
add package.xml to standard documents

rkent:
@tfoote [suggested](https://github.com/ros-infrastructure/rosdoc2/pull/114#issuecomment-2065437667) that package.xml be displayed. Here it is.


tfoote (approved):
This is great to add to the standard docs. I think that getting most of the information onto the landing page will also be valuable eventually, but we may want to think a bit more holistically about this especially with cross referencing.

PR #116, opened 2024-04-18
Title change to ROS2 Package <thepackage>

rkent:
Would it be OK if we gave a slightly more descriptive title? IIUC correctly "Welcome to the documentation for ..." comes from the default package generation for sphinx. We have a little more information, like this is a ROS2 package we are showing. The title should say that. "Welcome to the documentation for ..." is pretty useless verbiage. When I come to the page, particularly if I have lots of tabs opened, it would be good to have more hints of what I am looking at.

We could mark the package title differently if you don't like the <thepackage> style. Italics would also be straightforward.

### DIFF
--- a/rosdoc2/verbs/build/builders/index.rst.jinja
+++ b/rosdoc2/verbs/build/builders/index.rst.jinja
@@ -1,7 +1,7 @@
 .. GENERATED_CONTENT by rosdoc2.verbs.build.builders.sphinx_builder.
 
-ROS2 Package <{{ package.name }}>
-=============={% for _ in package.name %}={% endfor %}=
+{{ package.name }}
+{% for _ in package.name %}={% endfor %}
 
 {{ package.description }}
 
--- a/rosdoc2/verbs/build/builders/index.rst.jinja
+++ b/rosdoc2/verbs/build/builders/index.rst.jinja
@@ -1,7 +1,7 @@
 .. GENERATED_CONTENT by rosdoc2.verbs.build.builders.sphinx_builder.
 
-Welcome to the documentation for {{ package.name }}
-================================={% for _ in package.name %}={% endfor %}
+ROS2 Package <{{ package.name }}>
+=============={% for _ in package.name %}={% endfor %}=
 
 {{ package.description }}
 


rosdoc2/verbs/build/builders/index.rst.jinja
```diff
@@ -1,7 +1,7 @@
 .. GENERATED_CONTENT by rosdoc2.verbs.build.builders.sphinx_builder.
 
-Welcome to the documentation for {{ package.name }}
-================================={% for _ in package.name %}={% endfor %}
+ROS2 Package <{{ package.name }}>
+=============={% for _ in package.name %}={% endfor %}=
```
christophebedard, line 4: :grin: 

```suggestion
ROS 2 Package <{{ package.name }}>
==============={% for _ in package.name %}={% endfor %}=
```

clalancette: Yes, definitely.

But I might argue that we don't want to call these "ROS 2 packages" at all.  For instance, Gazebo and Fast-DDS both have package.xml files, but I wouldn't necessarily consider either of them to be a "ROS 2 package".

How about "Package documentation for {{package.name}}", or something like that?

rkent: When rosdoc2 is used, aren't we presenting the documentation AS a ROS 2 package? Isn't gazebo_ros, when documented by rosdoc2, a representation of gazebo as a ROS 2 package?

I'm speaking as an amateur to experts here, so I really don't know the right answer, but I will argue for being as specific as reasonably possible. 'package' is such a generic term that "Package documentation for {{package.name}}" isn't saying much.

tfoote: For simplifying this I would actually suggest that we drop this down to just the package name. It ends up in several places you can see it here in the headline and the top navigation and the tab heading. 

image

If we don't put the package name at the front the tab becomes almost unusable. 

And re calling it a ROS package or not. That is moderately implicit based on where it's being documented or used. Like we don't put ROS in the package name. The fact that it's on a ROS website or from a ROS workspace that's enough for people to know that it's a ROS package. And we have written this documentation engine to be actually generic so it can be used for gazebo packages. Or potentially any package outside of the ecosystem if they want to use our standards.

tfoote:
```suggestion
<{{ package.name }}>
={% for _ in package.name %}={% endfor %}=
```

tfoote: image
 simplified view with tab title, header all consistent

PR #118, opened 2024-04-19
Move Index to toc, remove Search link

### LINKED ISSUE #117: Include index in toc, not on home page
Per [here](https://stackoverflow.com/a/75972689) in sphinx 5.2 you can now easily include the index in the toc. Move it to there, and hide it from the home page.

As to the search page, this also can now be included in the toc, but as far as I can tell it has no value. Search is initialized by a text box at the top of the toc, and that goes to the Search page when entered. I just see a blank page when I click directly on Search. Am I missing something? If not, I'd remove the Search link completely.

rkent:
Fixes #117

Generally I'd like to remove items from the home page that are better accessed directly from the toc sidebar, this is one case of that. But in this case we are also adding something to the toc. Previously adding the index to the toc could only be done with a template, but from Sphinx 5.2 it is easily done.

The search link appears useless to me, search is accessed from the top of the sidebar.

### DIFF
--- a/rosdoc2/verbs/build/builders/index.rst.jinja
+++ b/rosdoc2/verbs/build/builders/index.rst.jinja
@@ -30,8 +30,8 @@ Welcome to the documentation for {{ package.name }}
 {% endif %}
 {% if has_readme %}.. include:: __readme_include.rst{% endif%}
 
-Indices and Search
-==================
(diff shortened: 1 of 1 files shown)


tfoote (approved):
This looks good to me

PR #119, opened 2024-04-19
Remove date from copyright

### LINKED ISSUE #77: Resolve what to print, if anything, for a copyright notice.
The copyright notice for packages is still at 2021, and flagged with a todo [here](https://github.com/ros-infrastructure/rosdoc2/blob/bccae77c43e6c6eff20c3aa1b2f77a9fd259c405/rosdoc2/verbs/build/builders/sphinx_builder.py#L229C1-L230C39).

```
# TODO(tfoote) The docs say year and author but we have this and it seems more relevant.
copyright = '2021, {package_licenses}'
```

IANAL but the subject of copyright notices has different opinions, but clearly this one (a fixed date) is not right. I would argue for:
- no date
- mention the license and authors from package.xml

Anyway this needs fixing.

rkent:
Fixes #77, see discussion there.


tfoote (approved):
Thanks this looks better.

PR #123, opened 2024-05-24
Support non-default python layouts

### LINKED ISSUE #122: Packages that claim python (ament_python or ament_cmake_python) fail if python non-standard and not specified
Lately I've been examining packages that fail to generate a rosdoc2 output. By far the most common cause of this is that the package claims to be python (either as ament_python or ament_cmake_python) but rosdoc2 fails to find the python package in a standard or specified (via rosdoc2.yaml) location.

Here's some notes of mine on failing python packages, and where the python is actually located:

camera_calibration: src/camera_calibration
mavros_extras: mostly c++, no python files, but has ament_cmake_python
message_filters: src/message_filters
microstrain_inertial_rqt: launch (this seems to be a launch helper, type ament_python)
nmea_navsat_driver: src/libnmea_navsat_driver (also ./launch)
nodl_python: src/nodl
ntrip_client: src/ntrip_client
python_orocos_kdl_vendor: no actual source. package has ament_cmake_python which triggers has_python
rmf_building_map_tools: many subdirectories, all under package.xml. So technically this is ./?
rmf_building_map_assets: has no python, but is type ament_python
rmf_traffic_editor_assets: has no python, but is type ament_python
rosbags_storage_mcap: has ament_cmake_python, c++ in src/, python in ros2bag_mcap_cli
rosbag2_storage_sqlite: has ament_cmake_python, python in rosbag2_sqlite3_cli
rqt: is a meta package, but has an empty resource/ directory. Type ament_python
rqt_action: source in src/rqt_action
rqt_bag: source in src/rqt_bag
... this seems to be the case for all of the rqt_* packages.
system_fingerprint: source in src/system_fingerprint
teleop_twist_keyboard: source in ./ that is at same level as package.xml
urdfdom_py: source in src/urdf_parser_py (that is, slightly different name than the actual package name.
vision_msgs_rviz_plugins: ament_cmake_python, mostly c++ but has samples/ directory with python.
webots_ros2_tiago: mostl resources, ament_python, but has python in launch/ directory.

Because rosdoc2 now includes documentation beyond just the code API, I don't think a complete failure is the right response to a failure to find python when expected. Also, many of these packages are using python src layout rather than flat. I think we should automatically search for python in that location as well as the standard location.

As for those that follow neither convention, it would be possible to search for python, but right now I'm thinking the right response is to just generate a warning, and proceed without generating a python API. The package author has the option of specifying the python source location using rosdoc2.yaml if they want.

rkent:
Fixes #122 

The changes made here are:

- a failure with the python source directory generates a warning instead of aborting
- if python source directory is not specified, then we search for it using flat layout and source layout. We prefer to find a python package with the same name as the ros package. But if not found by that name, we search for a single python package, and use that.

Tests are included for a variety of cases.

With these changes, the following packages that were not generating output before, now generate a proper output:

```
mavros_extras
message_filters
microstrain_inertial_rqt
nmea_navsat_driver
nodl_python
ntrip_client
python_orocos_kdl_vendor
rmf_traffic_editor_assets
rosbag2_storage_mcap
rosbag2_storage_sqlite3
rqt, plus many related packages rqt_*
system_fingerprint
webots_ros2_tiago
```

These packages that previously failed, now generate html but do not fully document their python:
```
rmf_building_map_tools: Succeeds, but has many subdirectories that are packages that it does not parse
teleop_twist_keyboard: Succeeds, but does not show python API for main file at ./ level.
urdfdom_py: Succeeds, but error in apidoc gives incomplete python API
vision_msgs_rviz_plugins: Succeeds, there is a samples/ python directory that is not parsed
```
One python package still fails for another reason:
```
camera_calibration: now has Could not import extension sphinx.ext.pngmath (exception: No module named 'sphinx.ext.pngmath'). Fails
```

Many of these could have been made to work previously with the correct specification in rosdoc2.yaml. But philosophically, I think we should try to succeed (with warnings) rather than just abort if the user fails to follow the conventions we expect.

### DIFF
--- a/rosdoc2/verbs/build/inspect_package_for_settings.py
+++ b/rosdoc2/verbs/build/inspect_package_for_settings.py
@@ -49,7 +49,7 @@ settings:
     ## be assumed for 'sphinx-apidoc' invocation. The user can provide the path
     ## (relative to the 'package.xml' file) where the Python modules defined by this
     ## package are located.
-    python_source: '{package_name}'
+    # python_source: '{package_name}'
 
     ## This setting, if true, attempts to run `doxygen` and the `breathe`/`exhale`
     ## extensions to `sphinx` regardless of build type. This is most useful if the
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -385,23 +385,32 @@ class SphinxBuilder(Builder):
 
         package_xml_directory = os.path.dirname(self.build_context.package.filename)
         # If 'python_source' is specified, construct 'python_src_directory' from it
+        python_src_directory = None
         if self.build_context.python_source is not None:
             python_src_directory = \
                 os.path.abspath(
                     os.path.join(
                         package_xml_directory,
                         self.build_context.python_source))
-        # If not provided, try to find the python source directory
-        else:
-            package_list = setuptools.find_packages(where=package_xml_directory)
-            if self.build_context.package.name in package_list:
-                python_src_directory = \
-                    os.path.abspath(
-                        os.path.join(
-                            package_xml_directory,
-                            self.build_context.package.name))
-            else:
+            if not os.path.isdir(python_src_directory):
+                logger.warning(f'python_source specified as {self.build_context.python_source} '
+                               'is not a directory')
                 python_src_directory = None
+
+        # If not provided or invalid, try to find the python source directory
+        if not python_src_directory:
+            search_dirs = ['.', 'src']
+            for search_dir in search_dirs:
+                where = os.path.abspath(os.path.join(package_xml_directory, search_dir))
+                package_list = setuptools.find_packages(where=where)
+                if self.build_context.package.name in package_list:
+                    python_src_directory = \
+                        os.path.abspath(
+                            os.path.join(
+                                where,
+                                self.build_context.package.name))
+                if python_src_directory:
+                    break
 
         # We will ultimately run the sphinx project from a wrapped directory. Create it now,
         # so that we can put generated items there.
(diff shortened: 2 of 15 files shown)


rkent:
@tfoote are you going to be able to review this?

rosdoc2/verbs/build/inspect_package_for_settings.py
```diff
@@ -49,7 +49,7 @@
     ## be assumed for 'sphinx-apidoc' invocation. The user can provide the path
     ## (relative to the 'package.xml' file) where the Python modules defined by this
     ## package are located.
-    python_source: '{package_name}'
+    # python_source: '{package_name}'
```
tfoote, line 52: Can the above cases not be covered by setting this non-standard location in their config files? 

Implementing custom discovery logic seem fragile and like it will end up with a lot of maintenance in the longer term versus setting the non-standard location in the configuration for those handful of packages.  And we can make sure that we provide a useful error/warning such that developers can know to set this setting in their config file.

tfoote (commented):
I'm on paternity leave until mid July so will have high latency and minimal bandwidth available. I put some thoughts in on this, I'd like to understand a bit more about whether we should be adding complexity to support automatically discovering the non-standard layouts/naming conventions vs providing a mechanism to just set this in rosdoc2.yaml etc.

rosdoc2/verbs/build/inspect_package_for_settings.py
rkent: Just to be clear, the reason for the change in the line that you quoted is because {package_name} is the default. I would expect that a sample config file shows what the default is, but does not set it unless setting it is required. It is not.

rkent:
Concerning "whether we should be adding complexity to support automatically discovering the non-standard layouts/naming conventions vs providing a mechanism to just set this in rosdoc2.yaml":

I feel strongly about:
- rosdoc2 should not crash (fail to generate any documentation) if the python source cannot be found.
- we should support by default both flat layout (as recommended) as well as src layout. (See the whole rqt_ universe of packages, like rqt_py_common for examples of src layout that I think we should support by default without a rosdoc2.yaml)

I don't feel strongly about fixing deviations from those two standards. One example is urdfdom_py, with the source in src/urdf_parser_py (that is, a slightly different name than the actual package name.)

But generally, I don't have a lot of confidence that package authors are going to add rosdoc2.yaml entries to fix minor differences in order to make rosdoc2 happy. I would rather have things "just work", and I am not convinced this adds much complexity to rosdoc2.

But I'll do what you want. If you want me to eliminate the search for misnamed subdirectories then I will.

nuclearsandwich:
> * rosdoc2 should not crash (fail to generate any documentation) if the python source cannot be found.

I agree with this first bullet. Expected error cases should be handled with error messages addressing the user rather than a raw stacktrace.

> * we should support by default both flat layout (as recommended) as well as src layout.

I have less of an opinion about this bullet but I'm not opposed to it either.

> But generally, I don't have a lot of confidence that package authors are going to add rosdoc2.yaml entries to fix minor differences in order to make rosdoc2 happy. I would rather have things "just work", and I am not convinced this adds much complexity to rosdoc2.

There's a pattern that I've seen for situations like this where a program is guessing about intended behavior which is to nag / warn that the program is guessing, then do it anyway, but suggest that a configuration be set to make what the program is guessing explicit. I think it's reasonable to push packages that don't adhere to standards to declare what they want instead in order to prevent the heuristics, which may be simple today, from growning unbounded without warning people that they might break under future versions without a config.

> I'll do what you want. If you want me to eliminate the search for misnamed subdirectories then I will.

I can't speak for @tfoote but splitting this PR in half, with the first adding support for the two agreed upon cases (flat and src) and providing a follow-up for others would allow me to feel comfortable reviewing and merging the former without waiting for @tfoote.

rkent:
This is the 'first half' that supports src layout. Also does not crash if there is an issue.

The earlier pushes had some git merge complications, this version should be a squash that just shows the current changes.

nuclearsandwich (approved):

PR #130, opened 2024-07-11
Update install instructions.

nuclearsandwich:
These README changes won't be valid until #129 is merged and python3-rosdoc2 is available in the ROS repositories.

📜 [Rendered](https://github.com/ros-infrastructure/rosdoc2/blob/install-instructions/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,17 +52,23 @@ Additionally, if you have a Python API then you will want to provide a Sphinx `c
 apt install -y python3-rosdoc2
 ```
 
+#### Installation from PyPI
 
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
FYI @rkent this is the last change I have planned for the 0.2.0 release to the ROS apt repositories and PyPI. I don't think that this PR needs to be in the release per se. But once the package is available I'd like to merge this.

cottsay (approved):

nuclearsandwich:
rosdoc2 0.2.1 is now available in the ROS repositories.

PR #132, opened 2024-07-12
Fix graphviz typo in stdeb.cfg.

nuclearsandwich:
Of course, discovered after publishing to apt and PyPI (which is unaffected by this issue).
I thought I'd done a successful dry-run but clearly that was before I snuck the last commit on #128 to include "grapviz". 🤦🏼


nuclearsandwich:
@rkent just to prove I was serious about being able to do frequent rosdoc2 releases. I'll do two in a row. 😄😢

cottsay (approved):

PR #135, opened 2024-07-22
Use jazzy, iron, and rolling in tests

rkent:
Welcome back, @tfoote ! 

This just updates our test matrix to include jazzy. I keep iron because IIUC it is running on older ubuntu and python, so might have different issues.


tfoote (approved):
Thanks! It was a good break, but I'm glad to be back too. 

This looks good. I switched the order to keep it alphabetized. But it's good to have coverage now that jazzy is out.

PR #136, opened 2024-07-25
Prevent package finders from processing the wrapped directory

### LINKED ISSUE #125: Add COLCON_IGNORE to the generated documentation?
I don't know if it is a wrong usage I make but the created documentation files appear in my ros2_workspace. When I try to colcon build again it complains about duplicate packages names since they are also found in the documentation.

A simple COLCON_IGNORE file would avoid that I think.

rkent:
Fixes #125 

Adds a blank COLCON_IGNORE file to the wrapped directory, to prevent package finders from thinking this is a package to process, as suggested by @damien-robotsix in #125

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -16,6 +16,7 @@ from importlib import resources
 import json
 import logging
 import os
+from pathlib import Path
 import shutil
 import subprocess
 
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
I'm general we want to discourage having the content in the source, but in the workspace makes sense. So avoiding the colcon walk makes sense. One request for the code.

tfoote (approved):
Thanks for the iteration. lgtm

PR #139, opened 2024-08-07
Add scan option to build packages in subdirectories

rkent:
Replaces #81

This PR implements a new top-level option, 'scan', which scans subdirectories, and builds any found packages. It uses the same options as ```build``` with a couple of additions: a per-package timeout, and a maximum number of packages.

I've been running variations of this for a few weeks now in a nightly Jenkins build of all ros2:iron packages, so the available output has been tweaked to focus on that environment.

It captures output from each package build in a file stdout.txt in the docs_build folder. Summary information is printed in the more typical console locations. Looking at the notes from #81, I believe this implements most of those requests, with the exception of the request to fail on first error. Plenty of existing packages have errors, so I don't really see the value of this if you are running a full distro. If you are only running a repository, it runs quickly enough that I don't see the value of that option either. I do include though a maximum number of packages options which is useful to limit time when testing. If you really want that feature, let's do that as a followup, it would not be very hard. But please state the use case.

On my system, if I set a ```-t 1``` timeout then about a third of the test packages time out, which can give some idea of what error output looks like. So I can run ```rosdoc2 scan -p test -t 1``` from the rosdoc2 directory as a simple test. Or you could do that on a repository with multiple packages instead.

I used colcon as requested, but I have some comments on that. First, the documentation on colcon that I could find is all from the perspective of using the resulting commands that are generated using colcon. I could find no hints of how to use it as a python library. Second, the code has multiple levels of abstraction, with key features being added through extensions. Examining the code for, say, colcon_core.package_discovery give no clue on what options are supported, as all of that is added in extensions. What extensions? I have no idea. Basically through trial and error and guesswork, I was able to discover that colcon-common-extensions give the functionality that I need, with ```--base-paths``` as the option I needed to set. Third, using colcon as a package discovery library requires 7 lines of obscure setup code. Yes this was partly written as a demo for me, but this is not at all the expectation one has when using library code. In summary, I would not recommend to anyone else that they attempt to use colcon as library code like this.


rkent:
Here is the list of failed packages from a recent ros2:iron run of this with the default 15 minute timeout (this is a 10 year old i7 machine using a physical hard disk):
```
37 packages failed:
rviz_common: retval=1: RuntimeError Sphinx-build exited with return code '2'
azure_iot_sdks: retval=1: RuntimeError Failed to find a ROS package at given path 'rosindex/_plugins_data/checkout/azure-iot-sdk-c/github-Azure-azure-iot-sdk-c/azure-iot-sdk-c'
ecl_geometry: retval=1: RuntimeError Sphinx-build exited with return code '2'
rotconv: retval=1: RuntimeError Failed to find a ROS package at given path 'rosindex/_plugins_data/checkout/rot_conv_lib/github-AIS-Bonn-rot_conv_lib/rot_conv_lib'
proxsuite: retval=2: TimeoutError runtime 900.351 seconds
gazebo_plugins: retval=1: RuntimeError Sphinx-build exited with return code '2'
magic_enum: retval=1: RuntimeError Sphinx-build exited with return code '2'
avt_vimba_camera: retval=1: RuntimeError Sphinx-build exited with return code '1'
fastcdr: retval=1: RuntimeError Failed to find a ROS package at given path 'rosindex/_plugins_data/checkout/fastcdr/github-eProsima-Fast-CDR/fastcdr'
octomap-distribution: retval=1: RuntimeError Failed to find a ROS package at given path 'rosindex/_plugins_data/checkout/octomap/github-octomap-octomap/octomap'
mrpt2: retval=1: RuntimeError Sphinx-build exited with return code '2'
libcaer: retval=2: TimeoutError runtime 900.348 seconds
mrpt_path_planning: retval=1: RuntimeError Sphinx-build exited with return code '2'
vitis_common: retval=2: TimeoutError runtime 900.349 seconds
sm_multi_stage_1: retval=2: TimeoutError runtime 900.349 seconds
sm_pack_ml: retval=2: TimeoutError runtime 900.349 seconds
VRPN: retval=1: RuntimeError Failed to find a ROS package at given path 'rosindex/_plugins_data/checkout/vrpn/github-vrpn-vrpn/vrpn'
VISP: retval=1: RuntimeError Failed to find a ROS package at given path 'rosindex/_plugins_data/checkout/visp/github-lagadic-visp/visp'
urdfdom_headers: retval=1: RuntimeError Failed to find a ROS package at given path 'rosindex/_plugins_data/checkout/urdfdom_headers/github-ros-urdfdom_headers/urdfdom_headers'
nav2_costmap_2d: retval=1: RuntimeError Sphinx-build exited with return code '2'
ecl_containers: retval=1: RuntimeError Sphinx-build exited with return code '2'
sick_scan_xd: retval=2: TimeoutError runtime 900.347 seconds
robot_localization: retval=1: RuntimeError Sphinx-build exited with return code '2'
naoqi_driver: retval=1: RuntimeError Sphinx-build exited with return code '2'
ecl_manipulators: retval=1: RuntimeError Sphinx-build exited with return code '2'
pinocchio: retval=2: TimeoutError runtime 900.351 seconds
ecl_converters: retval=1: RuntimeError Sphinx-build exited with return code '2'
fastrtps: retval=2: TimeoutError runtime 900.357 seconds
geodesy: retval=1: RuntimeError Sphinx-build exited with return code '2'
ign_rviz: retval=1: RuntimeError Sphinx-build exited with return code '2'
cv_bridge: retval=1: RuntimeError Sphinx-build exited with return code '2'
camera_calibration: retval=1: RuntimeError Sphinx-build exited with return code '2'
transmission_interface: retval=1: RuntimeError Sphinx-build exited with return code '1'
ur_client_library: retval=1: RuntimeError Sphinx-build exited with return code '2'
IFM3D: retval=1: RuntimeError Failed to find a ROS package at given path 'rosindex/_plugins_data/checkout/ifm3d_core/github-ifm-ifm3d/ifm3d_core'
ign_rviz_plugins: retval=1: RuntimeError Sphinx-build exited with return code '2'
rmf_visualization_schedule: retval=1: RuntimeError Sphinx-build exited with return code '2'
```
I can look at the logs on docs_build, for example, for rviz_common to see that the error was:
```
Exception occurred:
  File "/rosdoc2/venv/lib/python3.10/site-packages/breathe/renderer/sphinxrenderer.py", line 334, in create
    cls, name = DomainDirectiveFactory.cpp_classes[args[0]]  # type: ignore
KeyError: 'property'
```

During the active run, the console (Jenkins) output looks like this excerpt:
```
21:53:08 Begin processing geometry2
21:53:13 Package joint_state_publisher_gui succeeded in 10.981 seconds
[rosdoc2_scan] [INFO] joint_state_publisher_gui successful (1412/1440)
21:53:13 Begin processing system_modes_msgs
21:53:17 Package geometry2 succeeded in 8.858 seconds
[rosdoc2_scan] [INFO] geometry2 successful (1413/1440)
21:53:17 Begin processing mola_test_datasets
21:53:31 Package system_modes_msgs succeeded in 17.519 seconds
[rosdoc2_scan] [INFO] system_modes_msgs successful (1414/1440)
21:53:31 Begin processing ament_cmake_test
21:53:32 Package mola_test_datasets succeeded in 14.247 seconds
[rosdoc2_scan] [INFO] mola_test_datasets successful (1415/1440)
```

rkent:
A few notes on the code:

- I had to abandon running sphinx-build directly as a python module rather than in a subprocess, because exhale when it dies does an abrupt system exit without cleanup, which tends to hang the process pool since I cannot catch it.
- I replaced the pipx tests with a venv test. There does not seem to be any interest in promoting pipx in the ros2 community, while venv is important for development, and the install requirements can be different if not starting from a ros2 enabled system.

rkent:
I've done additional work on the whole area of locating packages to document in the last couple of weeks. I am now firmly convinced that colcon is the wrong tool for this use in rosdoc2, and I have [replaced it](https://github.com/rkent/rosdoc2/commit/b47e3a6f699618bc0d9a613a5d131327fec66588) with catkin's ```find_packages_allowing_duplicates``` in my fork. I'll eventually update this PR with that change, but for now please hold this PR.

Here's some background. colcon seems to be designed to find packages for building. Fine, but that is not what we are doing. Specifically (in addition to the complications mentioned above) it additionally scans CMakeLists.txt to try to find what it thinks is a package. Fine for building, not good for us since we need a real package.xml to process. So colcon is finding packages that we can't actually process, so it just ends up causing errors (like the ```urdfdom_headers: retval=1: RuntimeError Failed to find a ROS package at given path 'rosindex/_plugins_data/checkout/urdfdom_headers/github-ros-urdfdom_headers/urdfdom_headers'``` above).

I've fixed that in [my fork](http://rosdoc2.rosdabbler.com/urdfdom_headers/), but that fix (which would partially be at the buildfarm level) needs more tweaking. In any case, you have to replace colcon with catkin to eliminate nuisance errors.

rkent:
OK this is the updated version, reverting to catkin from colcon. Reasons are above.

Yes seems like a big PR, but a lot of this is just rearranging the tests so I can reuse code.

rosdoc2/verbs/build/builders/doxygen_builder.py
```diff
@@ -243,7 +244,8 @@ def build(self, *, doc_build_folder, output_staging_directory):
         logger.info(
             f"Running Doxygen: '{' '.join(cmd)}' in '{working_directory}'"
         )
-        completed_process = subprocess.run(cmd, cwd=working_directory)
+        completed_process = subprocess.run(
+            cmd, cwd=working_directory, stdout=sys.stdout, stderr=sys.stderr)
```
tfoote, line 248: Isn't this doing the default behavior with more complexity. The default is that subprocess.run doesn't capture the output which means it goes to the console. 

```
>>> import sys
>>> subprocess.run('echo hi', shell=True, stdout=sys.stdout, stderr=sys.stderr)
hi
CompletedProcess(args='echo hi', returncode=0)
>>> subprocess.run('echo hi', shell=True)
hi
CompletedProcess(args='echo hi', returncode=0)
```

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -523,7 +523,8 @@ def build(self, *, doc_build_folder, output_staging_directory):
                 logger.info(
                     f"Running sphinx-apidoc: '{' '.join(cmd)}' in '{wrapped_sphinx_directory}'"
                 )
-                completed_process = subprocess.run(cmd, cwd=wrapped_sphinx_directory)
+                completed_process = subprocess.run(cmd, cwd=wrapped_sphinx_directory,
+                                                   stdout=sys.stdout, stderr=sys.stderr)
```
tfoote, line 527: Same behavior here looks the same as default.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -533,12 +534,18 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
+        ]
         logger.info(
-            f"Running sphinx_build with: [{wrapped_sphinx_directory}, '{sphinx_output_dir}]'"
+            f"Running Sphinx-build: '{' '.join(cmd)}' in '{wrapped_sphinx_directory}'"
         )
-        returncode = sphinx_main([wrapped_sphinx_directory, sphinx_output_dir])
-        msg = f"sphinx_build exited with return code '{returncode}'"
-        if returncode == 0:
+        completed_process = subprocess.run(cmd, cwd=wrapped_sphinx_directory,
+                                           stdout=sys.stdout, stderr=sys.stderr)
```
tfoote, line 546: uncaptured here too

rosdoc2/verbs/scan/impl.py
```diff
@@ -0,0 +1,217 @@
...
+
+class Struct:
+    """Wrap argparse options to allow copies."""
+
+    def __init__(self, **entries):
+        """Create a dictionary from option entries."""
+        self.__dict__.update(entries)
+
+
+def prepare_arguments(parser):
```
tfoote, line 58: It would be good to expose BATCH_SIZE here too otherwise it's a magic number.

rosdoc2/verbs/scan/impl.py
```diff
@@ -0,0 +1,217 @@
...
+        batch_packages.clear()
+        packages = list(packages.values())
+        for i in range(len(packages)):
+            batch_packages.append(packages.pop())
+            if len(batch_packages) >= BATCH_SIZE:
+                break
+        logger_scan.info(f'Begin batch # {batch_no}')
+        for package in batch_packages:
+            logger_scan.info(f'Adding {package.name} for processing')
+        pool = mp.Pool(maxtasksperchild=1)
```
tfoote, line 116: We should definitely allow the ability to control the pool size as an argument. Otherwise some machines could easily be overrun if their memory to CPU core counts are different than we usually use.

rosdoc2/verbs/scan/impl.py
```diff
@@ -0,0 +1,217 @@
...
+    batch_packages = []
+    batch_no = 0
+    failed_packages = []
+    while len(packages) > 0:
+        batch_no += 1
+        batch_packages.clear()
+        packages = list(packages.values())
+        for i in range(len(packages)):
+            batch_packages.append(packages.pop())
+            if len(batch_packages) >= BATCH_SIZE:
```
tfoote, line 111: At a broader view, what is the value of the batching for running this? It basically forces the mp.Pool to be reset periodically. And with the current default it's not expected to ever be hit. As such might it make sense to reduce the complexity and not batch things before the pool. If someone wants to debug a build/scan they can change the argument to a specific subfolder which is where the problem is being encountered more intuitively than trying to tune the batch size.

test/utils.py
```diff
@@ -0,0 +1,187 @@
+# Copyright 2024 R. Kent James <kent@caspia.com>
```
tfoote, line 1: This file looks like the content is primarily moved content and not newly written so it should retain the original copyright owner.

tfoote (commented):
Overall adding the new verb is a good way to isolate the search capability. And it's calling through to the underlying build main which is good. 

I have a few smaller cleanups that would be great to resolve. I think some are development/debugging legacy that can probably be just cleaned out.

rosdoc2/verbs/build/builders/sphinx_builder.py
rkent: ditto

rosdoc2/verbs/scan/impl.py
rkent: The batching originally got added during development, when things were dying and it was really hard to figure out where and why. With small batch sizes, the problem could be isolated to a small number of possible packages to look for problems. As a debugging tool, it was easy enough to just change the value in code to enable batching. I don't really see the value except in debugging, hence it was not exposed to the user.

I have not had issues lately with it dying though, so I think I've caught all of the problems with other tools (adding a timeout, running Sphinx as a subprocess again, better logging of individual package outputs.) But I left it in since it was useful during initial debugging. But I don't feel strongly either, as --max-packages and --timeout mostly serve that need. I think I would rather remove it than expose it to the user though, as it really is not of any value except to the rosdoc2 developer. So unless you have another viewpoint, I'll remove the batching.

test/utils.py
rkent: OK

rosdoc2/verbs/build/builders/doxygen_builder.py
rkent: The reason that I did this is so that I can remap sys.stdout and sys.stderr to a file, and capture the output of individual package runs. IIRC from my tests, without this I could not remap to capture the output.

I was definitely following my "Programming is an Experimental Science"™ philosophy here though, trying things until something worked. But I'll look this over again, and confirm that I really need to do this. It seems strange to me as well that this is needed, but it seems to work.

rosdoc2/verbs/scan/impl.py
rkent: As discussed below, I don't think this is useful to the typical user, so I'll just remove batching instead.

rosdoc2/verbs/scan/impl.py
rkent: Wilco.

rosdoc2/verbs/build/builders/doxygen_builder.py
tfoote: If you do find that this helps with that extra bit of debuggability, it's fine to keep it. It would be great if you could add a comment to explain it so others can also find this benefit too.

rkent: I tried this again, and I can confirm that leaving ```completed_process = subprocess.run(cmd, cwd=working_directory)``` without adding ```, stdout=sys.stdout, stderr=sys.stderr``` does not work. The outputs go to the console, which is not what we want, we want to capture to a file.

I tried looking at pytest to see how they do it, but got lost in the levels of abstraction. Pytest has a lot of code around capturing, but I am sure they deal with a lot of edge cases that we don't care about.

If I look at subprocess, the code says:
```
def run(*popenargs,
        input=None, capture_output=False, timeout=None, check=False, **kwargs):
    """Run command with arguments and return a CompletedProcess instance.

    The returned instance will have attributes args, returncode, stdout and
    stderr. By default, stdout and stderr are not captured, and those attributes
    will be None. Pass stdout=PIPE and/or stderr=PIPE in order to capture them,
    or pass capture_output=True to capture both.
```
So in fact the default for syserr is not ```sys.stderr``` but None. We want to capture the output to our own PIPE, which we store in ```sys.stderr```.

We would not have to store that pipe in ```sys.stderr``` in the subprocess, we could define a variable for it,  but by doing so we also capture output generated by, for example, ```print(f"[rosdoc2] adding markdown parser", file=sys.stderr)```

rosdoc2/verbs/scan/impl.py
tfoote: Yeah, I think now that it's working well, it would be better to remove it and keep the much simpler flow of build and list and execute it in a pool.

rkent:
The commit I just did addresses all review comments, AFAICT.

rkent:
Re "If you do find that this helps with that extra bit of debuggability, it's fine to keep it"

If this comment is about the ```stdout=sys.stdout, stderr=sys.stderr``` then no, this is not about debugging, it is essential. As for a comment, if I found this strange I would look for the history, which will point here with a lot more detail than is likely in a comment. I can add a comment though if you like (or you could just add one when you land).

rosdoc2/verbs/build/builders/doxygen_builder.py
tfoote: Sorry I missed that you're changing the sys.stdout/err in the per package scan code here https://github.com/ros-infrastructure/rosdoc2/pull/139/files#diff-4d67c84fa514f2538371532c7f2d3979b42da8852649e61ed9cac143e83f1c55R169-R180

Which is capturing the output different than capturing it through the subprocess native mechanisms. Instead of changing sys.stderr at the global level, I think it would make sense to extend the main_impl to include the ability to pass an alternative output file. And if that output file is set, then pass that through to the subprocess invocation. This will avoid worries about race conditions on globally redirecting for potential other implementations of parallel threading.

rkent: I'm having a hard time understanding and implementing what you want here.

First, I don't see any mechanism within Pool to control capture of stdout and stderr, unlike the basic subprocess.run(). Do you know how to do that?

Once I get into package_impl, output is going to the normal sys.stdout and sys.stderr, and we have to fix that (there is no subprocess.run involved with build_main_impl). I fixed if via ```sys.stdout = outfile```. I don't see what else to do.

I was uncertain if that was the correct approach, but then I ran across contextlib.redirect_stdout in the standard library, and they do exactly what process_impl does: save the old value of sys.stdout, change sys.stdout to the new value, then restore it when they are done.

So I need more clarity about what you want here.

tfoote:
What I was thinking is that you set 
`options.console_output_file_log = os.path.join(options.doc_build_directory, f'{package.name}.txt')` after outfile is created: https://github.com/ros-infrastructure/rosdoc2/pull/139/files#diff-4d67c84fa514f2538371532c7f2d3979b42da8852649e61ed9cac143e83f1c55R171

Then the output_file goes into `build_main_impl` via the options struct. 

Then in main_impl https://github.com/ros-infrastructure/rosdoc2/blob/24c51d595e4477f11b3929c19e2fb6031936f42e/rosdoc2/verbs/build/impl.py#L133

The options get passed into the build_context.tool_options: https://github.com/ros-infrastructure/rosdoc2/blob/24c51d595e4477f11b3929c19e2fb6031936f42e/rosdoc2/verbs/build/inspect_package_for_settings.py#L129 

So in the builder itself you can use a conditional. just above https://github.com/ros-infrastructure/rosdoc2/blob/24c51d595e4477f11b3929c19e2fb6031936f42e/rosdoc2/verbs/build/builders/doxygen_builder.py#L246

```
if build_context.tool_options.console_output_file_log:
    with open(build_context.tool_options.console_output_file_log, 'w') as fh_log:
         subprocess.run( ... stdout=fh_log, stderr=subprocess.PIPE)
```

This will allow you to not rely on running the builds in different processes via the multi process pool in future applications. And others can potentially leverage the redirection of the log files in other verbs. It also keeps the log file opening much closer to the invocation which is better for cleanup as well as understanding that it might be intercepted.

rkent:
I finally found some time to switch my mental context from rosindex/html/javascript/css back into rosdoc2/python to understand your comments.

The approach that you outline won't really work as specified, at least if I maintain the goal of capturing the full build output in the outfile, because much of the output does not go through a subprocess.run called from the builder. There are at least 3 subprocesses in play: 1) scan subprocess implemented by package_impl. The rosdoc2 python code runs in this subprocess. 2) the Doxygen subprocess, and 3) the Breathe subprocess. The output of all three needs to be mapped to the same outfile. I can't really open the scan subprocess with a remapped output file AFAICT, because that is opened by the Pool. So I still need to do the stdout & stderr remapping in the scan subprocess to capture, for example, simple print statement output.

I could move the remapping logic into build/impl.py, but it would essentially be the same logic currently in package_impl. That though has the serious downside that package_impl cannot easily write to the outfile if it is opened and closed in build_main_impl. I believe it is important for the final termination messages for a particular package build to be written to both the scan output (which typically shows in the Jenkins console output) as well as in the outfile for a particular package. That is the point of code like:
```python
    finally:
        sys.stdout = old_stdout
        sys.stderr = old_stderr
        elapsed_time = '{:.3f}'.format(time.time() - start)
        if return_value != 0:
            print(f'{_clocktime()} Package at {package_path} failed {return_value}: {message}',
                  flush=True)
            logger.error(f'Package at {package_path} failed {return_value}: {message}')
...
```
because the print goes to the Jenkins output, while the logger is still mapped to the package outfile. This is particularly important when the package build is terminated by the watchdog timer, as without this the package outfile will just terminate without explanation. It is also difficult to close the outfile if the package build is terminated by the watchdog timer.

So I will do this change if you insist, but I personally don't believe the benefits justify the downside.

tfoote (approved):
Yeah, I think it's a better abstraction, but propagating the logfile and wiring it into all the locations to enable the consistent logging between the different elements does look like it will make things much more complex so lets stick with the global redirect.

PR #140, opened 2024-08-08
Use null as default for sphinx_sourcedir

rkent:
This is a followup to #138. The original documentation was wrong, doc/ is not a reasonable default for sphinx_sourcedir, as specifying that has the side effect of replacing the standard rosdoc2 output. That 'side effect' is actually the intended main effect of sphinx_sourcedir. We will need a new variable for the common issue that user documentation is in an alternate folder (typically docs/) but we do not want to override the default rosdoc2 output. That is a followup need.


tfoote (approved):
Thanks for cleaning up this documentation. It definitely got outdated as things evolved.

PR #141, opened 2024-08-12
Remove any existing package doc_build_directory

rkent:
Strange things can happen if there are leftover files in the docs_build directory. I recently came across one: if there is a previous run, then index.rst will already exist, so it will not be updated with index.rst.jinja if anything changes.

I cannot think of any good reason to leave the old docs_builds for the particular package, so this PR erases it at the beginning of the build.


tfoote (approved):
This seems reasonable to me. If we're not getting appropriate speedups from caching (aka we're getting stale data) this is better. And I don't think that there's a case when someone should be putting content into this directory and expecting that it would persist.

PR #143, opened 2024-08-23
Fix and expand "open" command

rkent:
Starting to do general documentation, tried 'open' and found it was lacking and broken.

Here's my update. With this, you can call 'rosdoc2 open' in several ways:

- ```rosdoc2 open``` will open the docs_output directory, which will be a list of directories of package names
- ```rosdoc2 open other_output_directory``` will work with a non-default output_directory (or file)
- ```rosdoc2 open package_name``` will open the documentation for package_name in the default location


tfoote (approved):
Thanks for cleaning this up.

PR #146, opened 2024-09-01
Removing any unsupported extensions

rkent:
Several packages add unsupported extensions in their conf.py files. Then when sphinx runs, it dies and we get no output.

This adds a check in the wrapping conf.py for the extensions that will be available when rosdoc2 is run in standard environments, and deletes (with a warning) any unsupported extension.

The PR also modifies a test to add all of these (plus one unsupported) to one of the test packages, to make sure that we both succeed with all standard extensions added, and that it does not die with an unsupported extension.

I'm trying to find PRs I can add which do apply cleanly to main (this is one),  but really I need resolution of #139 to make progress.  Most of my other patches cause conflicts which, if I attempt to resolve, will cause my rkent repo to diverge from ros-infrastructure in ways that will make it increasingly hard to maintain my fork.


tfoote (approved):
Thanks not crashing on these is much better behavior.

PR #149, opened 2024-10-09
Show package index in links

rkent:
When I am working with packages, I often find myself going back and forth between rosindex and rosdoc2 representations. rosindex links to rosdoc2, this adds the reverse link from rosdoc2 back to rosindex for the package. It is a simple PR that does not conflict with others.


tfoote (approved):
Adding this will greatly increase the navigability for users. 

For unindexed packages it will be a 404 but I think that's fine it might help encourage people to get things indexed.

PR #150, opened 2024-10-28
Remove unneeded .doctrees from sphinx output

rkent:
When running rosdoc2 over entire distros (using the scan option), disk usage is quite large. Investigating, the majority of that disk usage is in the .doctrees folder. AFAICT, this folder is only used during generation of the toctrees, and [is not needed for display of the documentation from the html](https://stackoverflow.com/questions/33904042/is-doctrees-folder-required-for-displaying-html-docs-with-sphinx).

This is also an issue on the standard buildfarm, though not as bad. See https://docs.ros.org/en/rolling/p/rclcpp/.doctrees/ for example.

This PR deletes the .doctrees folder after the sphinx documentation generation is complete.


tfoote (approved):
Confirmed as unnecessary https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-d

Issue #120, opened 2024-04-22
Packages that define conf.py in standrd locations usually fail


rkent:
I scanned rolling packages, and looked for packages with a conf.py in a standard location: doc/ or doc/source/. I found 18. Of these, 10 had serious failures (either no rosdoc2 result at all, or failure to show intended documentation:

conf.py failures:
1. geodesy
2. naoqi_driver
3. cv_bridge
4. mrpt_path_planning
5. robot_localization
6. ffmpeg_image_transport (readme_include fails)
7. ffmpeg_image_transport_tools (just shows package name)
8. quaternion_operation (just shows package name)
9. event_camera_py (rst files not shown; failed READMR include)
10. mrpt2

The reasons were varied, but some common reasons:
- error in user conf.py caused abort
- .rst files has ```.. include``` sections that referenced files not available in our wrapping

Of those who define conf.py, several would be better off with our newer rosdoc2 implementation rather than trying to roll their own. In others, the result would be better under our new *documentation* section rather than replacing all of the rosdoc2 output.

rkent:
Just as a FYI, I've been working recently on trying to understand and overcome issues of rosdoc2 failures from a variety of sources, including this one. So this issue is rather stale, but eventually I'll propose PRs for these issues.

PR #155, opened 2024-11-04
Set sys.path in wrapping conf.py so custom conf.py does not need to

rkent:
This PR came about because I was trying to figure out why launch did not generate any useful python documentation.

Setting the python directory is tricky for the user, combining both a wrapping directory as well as conf.py directory depth. No real reason they need to do that though, we can do it for them (but they could still override it in their conf.py if they want).

This does not fully solve the problem with launch, as they have another problem in their custom conf.py which I will address there in a PR. But this might help #120


rkent:
As I am looking at packages that are struggling with rosdoc2, I cam across another reason why this PR is important. The recommended way at the moment to set the python path in conf.py is with a complicated statement like:
```
sys.path.insert(0, os.path.abspath(os.path.join('..', '..', '..', '..', 'cv_bridge', 'python')))
```
The problem though is that this depends on where rosdoc2 is run from. If you run it from the repo, or you run it from the package directory itself, changes what the above should be. That is clearly not something that is acceptable, so this needs to be set in the wrapping conf.py (as in this PR), not the user conf.py

rkent:
After more thinking about the sys.path issue in conf.py, I realized that there really is no reliable way to do that in conf.py. So I changed the code comment in a followup commit that discourages setting it there.

tfoote (approved):
Thanks this looks like a good way to avoid requiring user customization.

PR #159, opened 2024-11-18
Add never_run_doxygen and never_run_sphinx_apidoc options

rkent:
Some large packages generate API documentation themselves. These same packages often take large amounts of time in rosdoc2 runs, with output that is not useful since the documentation options are often not optimal compared to the externally generated documentation.

This PR allows rosdoc2.yaml options to disable attempts by rosdoc2 to generate API docs in those cases.

An example is pinocchio which has complete documentation, including API documentation, at https://gepettoweb.laas.fr/doc/stack-of-tasks/pinocchio/master/doxygen-html/


rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -542,9 +542,11 @@ def build(self, *, doc_build_folder, output_staging_directory):
...
 
         build_context = self.build_context
-        has_python = build_context.build_type == 'ament_python' or \
-            build_context.always_run_sphinx_apidoc or \
-            build_context.ament_cmake_python
+        has_python = (
+            build_context.build_type == 'ament_python'
+            or build_context.always_run_sphinx_apidoc
+            or build_context.ament_cmake_python) \
+            and not build_context.never_run_sphinx_apidoc
```
tfoote, line 549: Can we separate this out so that this condition triggers a helpful debug message like the doxygen one above?

tfoote (changes requested):
Adding these options makes sense. One request to improve the debug messages.

rosdoc2/verbs/build/builders/sphinx_builder.py
rkent: Done.

tfoote (approved):
Thanks for the iteration. It looks good to me.

PR #160, opened 2024-11-21
Fix link whitespace to remove warning

rkent:
Fixes a warning during rosdoc2 generation.


tfoote (approved):
Thanks for the cleanup

PR #161, opened 2024-11-21
Add repository url if missing

rkent:
Currently, very few package authors actually set the package repository url, which is an official link in package.xml  But in navigating documentation, it is very important to be able to move as smoothly as possible between the rosindex package entry, rosdoc2 entry, and the repo. The repo url is readily available in rosdistro, so this PR looks it up if missing, and adds it.

Note this requires that we know the currently active rosdistro, which is not something that can be discerned directly from the package source. So the PR looks for ROS_DISTRO in the environment, and uses that if found. I'm not sure if the build farm sets that or not for doc jobs, if not that will need to be done to support this.

This also adds a new level of complexity to rosdoc2, that it is now using external information downloaded from the internet instead of just the local repo information for its build. I have a certain reluctance to add that additional complexity, but I believe the link to the repo URL is really important in documentation, so I think it is worth adding this.

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -29,6 +29,7 @@ from ..collect_inventory_files import collect_inventory_files
 from ..create_format_map_from_package import create_format_map_from_package
 from ..generate_interface_docs import generate_interface_docs
 from ..include_user_docs import include_user_docs
+from ..package_repo_url import package_repo_url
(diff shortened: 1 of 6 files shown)


clalancette:
> I'm not sure if the build farm sets that or not for doc jobs, if not that will need to be done to support this.

It currently does not.

tfoote (approved):
Adding this here makes sense. It can hopefully fill in more by default. 

We should think more about having ways to help integrate the documentation experience across sites with improved cross referencing based on the distro.

PR #162, opened 2024-12-09
Interpret enable_breathe and enable_exhale as never_run_doxygen option.

rkent:
Many packages define enable_breathe=false and enable_exhale=false in rosdoc2.yaml, but that is not a valid option there. It IS a valid option in rosdoc2_settings in config.yaml. I suspect there is some cargo culting goin on here (or some incorrect tutorial). What I assume they intend from this is that they do not want to generate C++ API from a Doxygen run.

This PR captures those values, issues a deprecation warning, and then sets never_run_doxygen=true which we assume is the intended effect.

I'm not completely sure this is a good idea, but it is part of my overall effort to make rosdoc2 work as well as possible with the existing package repos. There are other ways to accomplish this (override in yaml extend, or just submit PRs to the affected repos). Opinions welcome.

Here is  a list of packages in rolling that do this:

  rclpy/rclpy      
  point_cloud_transport/point_cloud_transport      
  geometry2/tf2_ros_py      
  geometry2/tf2_ros      
  ros2_control/ros2_control      
  ffmpeg_image_transport_tools      
  image_pipeline/image_pipeline      
  flir_camera_driver/spinnaker_camera_driver      
  event_camera_py      
  ros2_controllers/ros2_controllers      
  foxglove_compressed_video_transport      
  ffmpeg_image_transport

### DIFF
--- a/test/packages/basic_cpp/Doxyfile
+++ b/test/packages/basic_cpp/Doxyfile
@@ -1,0 +1,21 @@
+# All settings not listed here will use the Doxygen default values.
+
+PROJECT_NAME           = "urdf"
+PROJECT_NUMBER         = ros2
+PROJECT_BRIEF          = "C++ parser for the Unified Robot Description Format (URDF)"
+
+INPUT                  = ./include
+RECURSIVE              = YES
+OUTPUT_DIRECTORY       = docs_output
+
+EXTRACT_ALL            = YES
+SORT_MEMBER_DOCS       = NO
+
+GENERATE_LATEX         = NO
+GENERATE_XML           = YES
+EXCLUDE_SYMBOLS        = detail details
+
+ENABLE_PREPROCESSING   = YES
+MACRO_EXPANSION        = YES
+EXPAND_ONLY_PREDEF     = YES
+PREDEFINED             += URDF_EXPORT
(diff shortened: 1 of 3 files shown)


test/packages/basic_cpp/Doxyfile
```diff
@@ -0,0 +1,21 @@
+# All settings not listed here will use the Doxygen default values.
```
tfoote, line 1: Did this leak in from another change?

tfoote (commented):
This is effectively a deprecation notice for what I think was an older removal of options, that likely was silent. 

Getting them building makes sense with the clear deprecation cycle. This should help anyone else who has private repos or unindexed ones. 

It would be great to open PRs on the indexed ones to suggest this change, but we don't need to follow up explicitly. 

One question about the Doxyfile

test/packages/basic_cpp/Doxyfile
rkent: No, but it is only weakly related to this PR. This PR deals with when Doxygen is run. I did not have any tests that explicitly added a Doxyfile, so I added one here to make sure that things still worked.

tfoote (approved):
Got it lgtm

PR #164, opened 2025-01-03
Support a custom index.rst.jinja

rkent:
We use a standard index.rst.jinja to specify the main output of rosdoc2. This PR allows the user to specify a custom version of that.


rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -727,15 +732,21 @@ def generate_wrapping_rosdoc2_sphinx_project_into_directory(
...
         package = self.build_context.package
         if not os.path.isfile(index_rst_path):
-            # Generate a default index.rst
-            logger.info('Using a default index.rst.jinja')
-            template_path = resources.files(
-                'rosdoc2.verbs.build.builders').joinpath('index.rst.jinja')
-            template_jinja = template_path.read_text()
-
+            # Did the user provide index.rst.jinja?
+            user_jinja_path = os.path.join(wrapped_sphinx_directory, 'index.rst.jinja')
```
tfoote, line 736: Since we're using the more modern pathlib via resources below this could be symmetric and the `read_text()` method could be pulled outside the if block and avoid the `with open ... read()` Then the if block is just picking the template path and printing to say which it will use.

tfoote (commented):
One suggested change, but otherwise this would be great to enable. 

And could we add a quick reference to this capability in the documentation. I'm not sure exactly where would be best. But at least mentioning the capability if people search for index.rst.jinja they'll get a hit.

rkent:
I added changes to address your issues.

Concerning documentation, I did a fairly comprehensive [documentation patch](https://github.com/rkent/rosdoc2/tree/major-documentation-rewrite) for rosdoc2 in my development branch, but that is 14 commits ahead of where we are now. At one point I had hoped that I could get all of my work on rosdoc2 landed fairly quickly, and we could point package authors to the "new" rosdoc2 and its documentation, but that does not seem to be possible. Anyway I pulled out the index.rst documentation from that, updated it, and landed it on this PR's branch.

tfoote (approved):
Thanks for pulling the docs out. I know it will cause extra effort to do the merge.

PR #166, opened 2025-01-17
Deprecate the --install-directory option

### LINKED ISSUE #165: Deprecate the `--install-directory` command line argument?
Hello,

While investigating the use of `rosdoc2` I stumbled upon the `--install-directory` command line argument and was trying to figure what it was used for / if it was actually useful to make use of it.

From what I see in the code itself, the argument do not seem to be used. This would be confirmed as well by the replies on https://github.com/ros-infrastructure/rosdoc2/issues/80 .

I see that the `--build-directory` is being deprecated, should the same be made for the `--install-directory`?
Or was the argument kept "in the case" it would one day be used?

Extra note, if done this would certainly close https://github.com/ros-infrastructure/rosdoc2/issues/14.

jeremieguichardpro:
resolves #165

### DIFF
--- a/rosdoc2/verbs/build/impl.py
+++ b/rosdoc2/verbs/build/impl.py
@@ -66,7 +66,7 @@ def prepare_arguments(parser):
     parser.add_argument(
         '--install-directory',
         '-i',
-        help='install directory of the package',
+        help='UNUSED, to be removed at some time (install directory of the package)',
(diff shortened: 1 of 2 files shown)


tfoote (approved):
Thanks it's good to deprecate this officially since we're not using it.

PR #168, opened 2025-02-05
Don't duplicate doc links if index.rst exists

rkent:
Currently, if there is an index.rst in a documentation folder, that does not override the standard directory search for documentation, which results in duplicate entries for documentation, or includes documentation folders that the user expected to be excluded since it was not included in index.rst

The heart of this PR got obscured by a lot of additional cleanup going on.

- renamed sphinx_project_directory to the more accurate rel_user_doc_directory
- reduced some of the levels shown in the documentation tree.  I don't believe in most cases it makes sense to show tree entries for the internals of documentation files.
- creation of a new test file to test the issue.

Really the heart of this is the code:
```python
        if 'index.rst' in files:
            # We assume that this index will also show any desired files in subdirectories
            dirs.clear()
```

You can see the point of this if desired by reverting the changes in `include_user_docs.py` and running the tests. The failed new test index_excudes_content  will show errors when files are shown that are not included in index.rst

### DIFF
--- a/rosdoc2/verbs/build/include_user_docs.py
+++ b/rosdoc2/verbs/build/include_user_docs.py
@@ -16,8 +16,6 @@ import logging
 import os
 import shutil
 
-from rosdoc2.slugify import slugify
-
 logger = logging.getLogger('rosdoc2')
 
 documentation_rst_template = """\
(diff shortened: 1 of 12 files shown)


tfoote (approved):
Thanks for the summary of the logic on this one. It makes sense to me to not do the default behavior if the user has replaced it.

PR #169, opened 2025-02-06
Create a links.rst file so links can appear in toc

rkent:
Most of the top-level items on the home page also appear in the table of contents, except for links. This adds those links to the TOC.


tfoote (approved):
These are great to add to the TOC

PR #171, opened 2025-02-06
Use real names for copied files and directories

rkent:
Early on, we decided that files copied into the staging folder would be given custom directory names, like `standard_docs` or `interfaces`. Sometimes though users try to link to documents or resources in those folders, which frequently fails. This partially corrects that, by using the real names of files and folders whenever possible.

Unfortunately there are many cases where this still does not work. Users are frequently using github to design their links, and github is more robust than simple sphinx, rewriting links to work with their website. See for example https://github.com/ros2/rcl_interfaces/tree/rolling/action_msgs where the readme has a link to a message, and to a Quality Declaration. Before this PR, both of those links do nothing (while they work in Github). After this, the .msg links will ask to download the .msg file, while the Quality Declaration link now properly links.

But I still think this is worth doing even though incomplete. There was no good reason to use artificial names in the first place.


tfoote (approved):
That's a great improvement to keep things consistent for the developers.

PR #173, opened 2025-02-19
Allow show_doxygen_html and disable_breathe options

rkent:
A number of packages are failing to generate documentation because breathe throws an error. Diagnosing those errors is tricky, and may not be worth the effort for some packages.

This PR adds two related options to rosdoc2, one to disable breathe, and another to show links to the Doxygen html.

Some example packages where this is an issue can be seen from [this link](https://github.com/rkent/docker_farm/blob/c17c33b69a2851ec7b62147f845aa7c28b24b684/common/rosdoc2/docs_ex.yaml#L13-L44) which shows a number of packages in my development build where these options are enabled. I believe in the [official api documentation](http://docs.ros.org/en/humble/p/) most or all of these not show any results, while for my [dev build](https://dev-rosdoc2.rosdabbler.com/humble/) the packages should be shown (except for magic_enum). For example, [here](https://dev-rosdoc2.rosdabbler.com/humble/ecl_geometry/) is the link for ecl_geometry, while the [official link](http://docs.ros.org/en/humble/p/ecl_geometry) is empty.

The link to the Doxygen html is not ideal, as it takes one away from the standard rosdoc2 API site layout. That's why we use breathe in the first place. But particularly for packages that have put effort into their Doxygen results, but cannot get breathe to work, this would be the best choice.


tfoote (approved):
Yeah, there's definitely significant value to unlock sites that have strong doxygen culture already that breathe can't process. Maybe in the future we can have a process that embeds the doxygen with our sidebar. Of course we don't want to reimplement breathe. But maybe it could be more of a wrapper that embeds the doxygen content. I hestitate to say it as it sounds hacky but maybe even a javascript or sed script that litterally injects the standard side bar into the generated content. 

But for now if we get more coverage I think that this is valuable. And we can meet existing projects with doxygen where they are at least. Not well integrated is better than not integrated. And we do still have the landing page from sphinx at least with all the standardized content.

ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/new-nav2-api-docs/49763/2

PR #174, opened 2025-02-20
Pin to sphinx < 8.2.0

cottsay:
There is a private API dependency on sphinx from breathe 4.35, and sphinx 8.2.0 changes that private API in a way that breaks breathe. This issue only appears to manifest in rosdoc2 jobs which process C++.

Example of this failure on the official buildfarm: https://build.ros2.org/view/Jdoc/job/Jdoc__rcl__ubuntu_noble_amd64/10/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     pyyaml
     rosdistro
     setuptools>=40.6.0
-    sphinx
+    sphinx<8.2.0
     sphinx-rtd-theme
     myst-parser
 packages = find:
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,6 +24,8 @@ long_description = file: README.md
 long_description_content_type = text/markdown
 keywords = rosdoc2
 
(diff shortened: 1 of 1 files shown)


tfoote:
Cross referencing there's an upstream fix already merged: https://github.com/breathe-doc/breathe/pull/1010#issuecomment-2671933577  but it needs a release: https://github.com/breathe-doc/breathe/issues/1022

tfoote:
The CI failure looks like it's related to #173 FYI @rkent

tfoote (approved):
This looks good. I added a comment and the error looks like it might be a flakey race condition. I'll ticket that separately.

tfoote:
The rebuild was successful: https://build.ros2.org/view/Jdoc/job/Jdoc__rcl__ubuntu_noble_amd64/11/

rkent:
> The CI failure looks like it's related to #173 FYI @rkent

Where are you seeing that?

PR #175, opened 2025-02-20
Shorten some long lines to please flake8

cottsay:


tfoote (approved):

tfoote:
Thanks for fixing this @cottsay do you know why #166 didn't get CI run on it? 

This looks like it should have been caught.

https://github.com/ros-infrastructure/rosdoc2/blob/ffee45dbe6e04f37e6a511c286c4a0ffa34c8306/.github/workflows/test.yml#L3-L7

cottsay:
> do you know why https://github.com/ros-infrastructure/rosdoc2/pull/166 didn't get CI run on it?

I have no idea. Everything seems configured correctly.

jeremieguichardpro:
@cottsay , @tfoote Sorry for not catching that, I should have run the linters on local...

Seeing that the issue was as well on some other lines I had not edited myself my best guess is that CI is disabled on PRs from forks. See the option that looks like:

image

tfoote:
Thanks that setting is an option for Private repositories, but it's not present for Public repositories. I think that this was a github hiccup on triggering.  Most of the recent pull requests are from external forks.

PR #177, opened 2025-02-20
Add '--version' option to print rosdoc2 version

christophebedard:
I wanted to check the version myself and saw there was no `--version` option. I thought this would be useful.

```console
$ rosdoc2 --help
usage: rosdoc2 [-h] [--version] [build | default_config | open | scan] ...

rosdoc2 builds documentation for ROS packages

options:
  -h, --help            show this help message and exit
  --version             show program's version number and exit

rosdoc2 command:
  Call `rosdoc2 [build | default_config | open | scan] -h` for help on a each verb.

  [build | default_config | open | scan]
$ rosdoc2 --version
rosdoc2 0.2.2.dev0
```

Note that `--version` can't be used with a verb:

```console
$ rosdoc2 build --help
usage: rosdoc2 build [-h] --package-path PACKAGE_PATH [--build-directory BUILD_DIRECTORY] [--install-directory INSTALL_DIRECTORY] [--cross-reference-directory CROSS_REFERENCE_DIRECTORY] [--base-url BASE_URL]
                     [--output-directory OUTPUT_DIRECTORY] [--doc-build-directory DOC_BUILD_DIRECTORY] [--debug] [--yaml-extend YAML_EXTEND]

Build the documentation of a ROS package

options:
  -h, --help            show this help message and exit
  --package-path PACKAGE_PATH, -p PACKAGE_PATH
                        path to the ROS package
  --build-directory BUILD_DIRECTORY, -b BUILD_DIRECTORY
                        UNUSED, to be removed at some time after September 1st, 2024
  --install-directory INSTALL_DIRECTORY, -i INSTALL_DIRECTORY
                        UNUSED, to be removed at some time (install directory of the package)
  --cross-reference-directory CROSS_REFERENCE_DIRECTORY, -c CROSS_REFERENCE_DIRECTORY
                        directory containing cross reference files, like tag files and inventory files
  --base-url BASE_URL, -u BASE_URL
                        The base url where the package docs will be hosted, used to configure tag files.
  --output-directory OUTPUT_DIRECTORY, -o OUTPUT_DIRECTORY
                        directory to output the documenation artifacts into
  --doc-build-directory DOC_BUILD_DIRECTORY, -d DOC_BUILD_DIRECTORY
                        directory to setup build prefix
  --debug               enable more output to debug problems
  --yaml-extend YAML_EXTEND, -y YAML_EXTEND
                        Extend rosdoc2.yaml
$ rosdoc2 build --version
usage: rosdoc2 build [-h] --package-path PACKAGE_PATH [--build-directory BUILD_DIRECTORY] [--install-directory INSTALL_DIRECTORY] [--cross-reference-directory CROSS_REFERENCE_DIRECTORY] [--base-url BASE_URL]
                     [--output-directory OUTPUT_DIRECTORY] [--doc-build-directory DOC_BUILD_DIRECTORY] [--debug] [--yaml-extend YAML_EXTEND]
rosdoc2 build: error: the following arguments are required: --package-path/-p
```

### DIFF
--- a/rosdoc2/main.py
+++ b/rosdoc2/main.py
@@ -19,6 +19,8 @@ from osrf_pycommon.cli_utils.verb_pattern import create_subparsers
 from osrf_pycommon.cli_utils.verb_pattern import list_verbs
 from osrf_pycommon.cli_utils.verb_pattern import split_arguments_by_verb
 
+from . import __version__ as VERSION
+
 COMMAND_NAME = 'rosdoc2'
 
 VERBS_ENTRY_POINT = f'{COMMAND_NAME}.verbs'
@@ -37,6 +39,7 @@ def main(sysargs=None):
     parser = argparse.ArgumentParser(
         description=f'{COMMAND_NAME} builds documentation for ROS packages'
     )
+    parser.add_argument('--version', action='version', version=f'{COMMAND_NAME} {VERSION}')
 
     # Generate a list of verbs available
     verbs = list_verbs(VERBS_ENTRY_POINT)
@@ -56,6 +59,10 @@ def main(sysargs=None):
     # Short circuit -h and --help
     if '-h' in pre_verb_args or '--help' in pre_verb_args:
         parser.print_help()
+        sys.exit(0)
+    # Short circuit --version
+    if '--version' in pre_verb_args:
+        parser.parse_args(sysargs)
         sys.exit(0)
 
     # Error on no verb provided


rosdoc2/main.py
```diff
@@ -37,6 +39,7 @@ def main(sysargs=None):
     parser = argparse.ArgumentParser(
         description=f'{COMMAND_NAME} builds documentation for ROS packages'
     )
+    parser.add_argument('--version', action='version', version=f'{COMMAND_NAME} {VERSION}')
```
christophebedard, line 42: Also, we can add `-v` as an alternative to `--version` if we want to.

rkent:
The current problem with this is that the version is not really very useful at the moment. The version was set with a PR that was preliminary to releasing rosdoc2 as a pip package, but that version has not been updated even though significant changes have been made to rosdoc2.

Since rosdoc2 buildfarm runs pull from the same branch that development PRs are landed to, the useful "version" would be the date (and perhaps commit tag) of the last patch.

I say this without having looked at the details of this PR.

christophebedard:
I agree that this is not helpful for doc jobs on the buildfarm.

I thought this would be useful when using `rosdoc2` locally. I believe I've installed `rosdoc2` using `apt` on my system (`python3-rosdoc2`, currently `0.2.1`), and also have a local clone of it somewhere (currently `0.2.2.dev0`). I just wanted to check which version I had in a terminal. There's other ways to find out, of course, but this is quick and easy (and is fairly typical).

Anyway, this is just a suggestion.

tfoote:
Adding this sounds like a good idea. I think it would also be helpful for it to query and print out the versions of the tools it depends on as well.

For example we just had the sphinx breathe conflict that this could help diagnose. But more likely it will be valuable to be able to detect missing dependencies.

tfoote:
Re the rolling versioning, we could get better about doing releases periodically or something else low overhead.

rosdoc2/main.py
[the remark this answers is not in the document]
rkent: All of the examples that I tried showed the command name with --version, so IMHO including COMMAND_NAME is best.

rkent: Yes please add the -v option, since that pattern has already been established with `-h, --help`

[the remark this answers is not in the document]
christophebedard: sounds good.

christophebedard:
> I think it would also be helpful for it to query and print out the versions of the tools it depends on as well.

Good idea. Should I do that in this PR? And should these be included in `-v,--version`, or a separate flag? Like `-vv,--version-deps`.

rosdoc2/main.py
```diff
@@ -19,6 +19,8 @@
 from osrf_pycommon.cli_utils.verb_pattern import list_verbs
 from osrf_pycommon.cli_utils.verb_pattern import split_arguments_by_verb
 
+from . import __version__ as VERSION
```
tfoote, line 22: As of 3.8 importlib.metadata provides a way to do this instead of the standalone file: https://docs.python.org/3/library/importlib.metadata.html#distribution-versions  I think that it would be a better way for maintainability and consistency.

tfoote (changes requested):
Lets get this in at first and we can do a follow up with the additional dependency versions. 

I think that importlib.metadata will also make that much easier too to get the dependency's versions.

rosdoc2/main.py
tfoote: As an example: https://stackoverflow.com/questions/15405636/pythons-argparse-to-show-programs-version-with-prog-and-version-string-formatt

tfoote (approved):

PR #178, opened 2025-02-21
Copy templates to actual sphinx project location

rkent:
Sphinx projects can include templates that modify how the output appears, for example adding content in a footer. This would typically by specified in a templates directory, relative to where conf.py exists.

The issue we face is that people who specify templates would typically do that in a doc/_templates directory, which is also where a conf.py file would be added (which we rename to __conf.py to avoid certain conflicts, or supply a default if missing).

However, the actual sphinx project that we run, using a conf.py that wraps either a user or a default conf.py, is run in the wrapped_sphinx_directory, not in wrapped_sphinx_directory/doc which is where the templates are. So a templates reference like '_templates' (which we define by default) is looking for the templates in wrapped_sphinx_directory/_templates, and not wrapped_sphinx_directory/doc/_templates where they are actually located.

This PR copies templates from where they exist (relative to the user's conf.py directory) to where they are needed (relative to the wrapped_sphinx_directory).

There are very few packages that are currently actually using templates. I did find one, ecal. Not a great example since they have their own complex doc generation, but they do have a template that defines a custom template. Without this patch, the footer does not get changed, but with it their custom footer gets used in all of the generated documentation.

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -49,12 +49,15 @@ rosdoc2_wrapping_conf_py_template = """\
 ## intersphinx mappings correctly, among other things.
 
 import os
+import shutil
 import sys
 if '{python_src_directory}' != 'None':
     sys.path.insert(0, os.path.abspath(os.path.join('{python_src_directory}', '..')))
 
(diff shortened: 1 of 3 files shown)


tfoote (commented):
This looks like a good solution

tfoote (approved):

PR #180, opened 2025-02-24
Revert "Pin to sphinx < 8.2.0 (#174)" as breathe has been updated.

rkent:
This reverts commit 95fa29de85a692ba9c312e70d8c67b912db9b47c.

Breathe has been updated now, and a more robust release procedure implemented in https://github.com/breathe-doc/breathe/pull/1023. I think we want to generally follow newer breathe and sphinx releases rather than pin to a fixed release.

I tested this, using a local venv, by first confirming that I could reproduce the original error by running rosdoc2 on jazzy rcl forcing breathe 4.35.0, then confirming that the error goes away with this PR.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,8 +24,6 @@ long_description = file: README.md
 long_description_content_type = text/markdown
 keywords = rosdoc2
 
-# Pin sphinx version under 8.2 until there's a fix for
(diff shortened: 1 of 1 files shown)


tfoote (approved):
It's great they got the release out quickly. We definitely want to keep up with new fixes. 

Thanks for seeing that and confirming the fix.

PR #183, opened 2025-02-27
Specify ROS_VERSION=2 to get correct build type

### LINKED ISSUE #181: Packages that are looking for "ROS_VERSION" fail
Some packages use conditional values in package.cml, like:

```xml
<export>
    <!-- We need catkin installed during builds so debian/rules finds /opt/*/setup.bash -->
    <build_type condition="$ROS_VERSION == 1">catkin</build_type>
    <build_type condition="$ROS_VERSION == 2">cmake</build_type>
</export>
```
rosdoc2 is ROS 2 only, but these packages are not getting the correct build_type, so Doxygen is not running.

One example is mp2p_icp (though there are other reasons there why Doxygen is not running).

rkent:
Fixes #181.

This affects many packages. One example is find_object_2d, which has no C++ api at http://docs.ros.org/en/rolling/p/find_object_2d/ but after this PR will have one.


tfoote (changes requested):
Most of our builds can assume the presence of ros_environment https://github.com/ros/ros_environment

We adding ROS_DISTRO parameterization a little while ago. 

We should set these two environment variables in the same way.  https://github.com/ros2/ros_buildfarm_config/pull/314

We could give a warning if we detect it's unset, but like the distro we shouldn't set this expected environment inside the tool.

rkent:
The difference here is that ROS_DISTRO changes, while ROS_VERSION is always '2' for rosdoc2.

I checked the documentation for os.environ before I did this, it says:

> We can also modify os.environ but any changes will be effective only for the current process where it was assigned and it will not change the value permanently.

I also checked running a python script locally that called os.environ, and the changed environment value did not persist in the console after python exited.

So I can try the change you suggested, but it seems to me to be a roundabout way of accomplishing this.

tfoote:
I have two reasons for this. First is to maintain a consistent path for the similar variables. If one is set internally but the other is has to be set externally it will be confusing to the user. We should improve the docs to recommend making sure that you've sourced the ROS workspace that you're planning to document before you run the build to get the appropriate. 

And secondly I'd like to leave the ROS_VERSION value not hard coded. We're not expecting to do a major version bump of ROS. But with ROS 1 sunset the ROS_VERSION value is going to not be very meaningful and we may want to leverage it for something else. And hard coding it in our tools will just risk future maintenance burdens. 

I've opened https://github.com/ros2/ros_buildfarm_config/pull/319 which should resolve it for the buildfarm. And for local reproduction I mentioned above about sourcing the ROS workspace which will give both the ROS_DISTRO and the ROS_VERSION so it's the same simple guidance w/o adding any complexity for the user.

rkent:
So as I understand this, we will not implement this PR here, but instead fix this on the build farm. I've added a note to myself to note in documentation the need to source the ROS workspace. I have a big PR coming with rewritten documentation.

PR #184, opened 2025-03-03
Allow user to specify index.rst in package directory

rkent:
Fixes https://github.com/rkent/rosdoc2/issues/4

We currently support an index.rst.jinja in the package directory, but I'd always intended to also support index.rst there. This fixes that.


tfoote (approved):
Adding support for the straight index.rst sounds good.

PR #185, opened 2025-03-04
Show rosdistro in TOC

rkent:
Since we now have ROS_DISTRO available, this shows the distro in the form '<package_name>: Rolling' following the format in the main ros documentation at https://docs.ros.org/en/rolling/


tfoote (approved):
That's great extra context.

PR #186, opened 2025-03-04
doc/index.rst should override glob of doc/ files

rkent:
Fixes https://github.com/rkent/rosdoc2/issues/41

I'm going through my list of issues, doing easy fixes.

If a user specifies an index.rst file in their doc folder, we should assume that contains everything needed to show the documentation in that folder. Currently, we are treating that like any other documentation file, that is it is included in a list of all of the other document files in that folder. With this change, if the index.rst file exists, only it is shown, and we assume that the user provided a link to whatever else they care about in that file.

### DIFF
--- a/rosdoc2/verbs/build/include_user_docs.py
+++ b/rosdoc2/verbs/build/include_user_docs.py
@@ -27,7 +27,7 @@ Documentation
    :titlesonly:
    :glob:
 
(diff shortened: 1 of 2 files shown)


tfoote (approved):
That makes more sense to follow their explicit direction if provided.

PR #188, opened 2025-03-04
Document testing

rkent:
This PR adds documentation for testing.

I want to start landing the documentation for the current version of rosdoc2. I wrote a bunch of  documentation several months ago in a branch https://github.com/rkent/rosdoc2/tree/major-documentation-rewrite. This is one file from that branch.

Is it OK if we land these one at a time (like in this PR), or would you rather bunch them into a big documentation PR?

I've been avoiding reviewing my documentation rewrite myself, while I found this one file fairly easy to test and add. So I would prefer one file at a time.

### DIFF
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -1,0 +1,22 @@
+# Testing
+
+To install rosdoc2 prerequisites for testing, run (preferably in a virtual environment)
+```bash
+pip install ---upgrade --editable .[test]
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Small batches is great. No need for a big monolith as long as it's generally not tightly coupled. It makes reviewing much easier.

PR #189, opened 2025-03-04
Document rosdoc2.yaml

rkent:
Here's documentation for rosdoc2.yaml

I modified an existing test package to add the suggested minimum rosdoc2.yaml there. No specific tests for that, mainly that it does not crash.


tfoote (approved):
This is great to get documented.

PR #192, opened 2025-03-08
Allow options for non-default extensions and themes

### LINKED ISSUE #191: Allow Custom Extensions in rosdoc2 for External Documentation Builds
## Description  
In [PR #146](https://github.com/ros-infrastructure/rosdoc2/pull/146), rosdoc2 removes unsupported extensions behind the scenes. While this behavior is beneficial for enforcing consistency in official ROS2 documentation, it introduces a problem for external users who rely on custom Sphinx extensions to maintain their own documentation standards.  

For instance, our organization uses a custom documentation tool, [adi_doctools](https://github.com/analogdevicesinc/doctools), to ensure a consistent experience and theme across all documentation builds. However, since `adi_doctools` is not an officially supported extension, it is automatically removed by rosdoc2, causing the build to fail when attempting to apply our theme.  

## Current Problem  
- The rosdoc2 API exposes a setting:  
  ```python
  rosdoc2_settings -> 'override_theme': False
  ```
- This suggests that a user can control whether to override the default theme.
- However, since rosdoc2 removes unsupported extensions, I cannot inject my custom theme (`adi_doctools`), and the build fails.
- When I configure `override_theme: False`, I expect that my extension remains loaded to apply the intended theme.
- Instead, the extension is removed, but the configured theme (`harmonic`) is still being referenced instead of the fallback `sphinx_rtd_theme`.

## Proposed Solution
Introduce an option to explicitly allow a user-defined set of extensions when running rosdoc2 outside of official ROS2 builds. This would differentiate:

- **Official ROS2 documentation builds** – where only approved themes and extensions are enforced.
- **External user builds** – where users can specify additional extensions like `adi_doctools` without being removed.

This could be achieved by:

- Allowing users to specify a list of extensions to be exempt from removal.
- Ensuring that when `override_theme: False` is set, a removed extension does not leave behind an incomplete theme configuration.

## Expected Behavior
- If `adi_doctools` is explicitly enabled in a non-ROS2 build, the extension should remain loaded.
- Official ROS2 builds can still enforce a restricted set of themes while allowing external users to configure their own themes without unexpected removals.

rkent:
Fixes #191 

Adds the new conf.py option: rosdoc2['allow_other_extensions']  If True, then rosdoc2 will include additional specified Sphinx extensions if they are installed.

With themes, supports the existing conf.py option: rosdoc2['override_theme']. If False, will now check if a specified theme is installed, and allow it if so. Otherwise, forces the default theme.

Changes are made to three tests, but only for visual confirmation (except for crashes) using the messages.

basic_cpp tests the case where we say we say we want to override extensions and theme, but the specified extension and theme do not exist, so the default is used anyway.

rclcpp tests the case where we specify valid non-standard extensions and themes, but they are removed/replaced because the appropriate settings in conf.py are not set.

src_python tests the case of successfully using a non-standard extension and theme.

In doing these, I discovered that the generated conf.py files were inconsistent in whether prints went to sys.stderr. I changed all to be consistent, removing the sys.stderr options in prints.

@Adrian-Stanea it would be good if you could review this, and approve if it fits your needs. Note I am not a committer, so a committer needs to look this over before it is accepted.


rkent:
Fixed a trailing whitespace. Not sure how that snuck in!

Adrian-Stanea:
I have tested the changes locally and can confirm that the revised logic for loading extensions, when present in the environment along with the `allow_other_extensions` option, meets my requirements.

From my tests:

* **Case 1:**
  *  `'allow_other_extensions': True`
  * Non-default extension listed in the extensions list
  * Extension **available** in the Python environment
  * Outcome: rosdoc2 correctly identifies and loads my extension, building the documentation with my custom theme set in the html_theme variable. This covers the scenario where external users rely on rosdoc2 for building customized documentation.
  
 * **Case 2:**
   * `'allow_other_extensions': True`
   * Non-default extension listed in the extensions list
   * Extension **not available** in the Python environment
   * Outcome: rosdoc2 recognizes that my extension is not present in the environment and cannot be imported, so it is omitted. The logic properly overrides the html_theme variable to the official ROS theme. This ensures that builds made in the build farm follow the predefined theme.

Adrian-Stanea (approved):

rkent:
Thanks for the review. The committer on this repo, @tfoote is away for three weeks, so this might have to wait that long. That does not mean it is rejected or unwanted.

tfoote (approved):
Thanks for the concrete requirements and proposed mechanism to maintain backwards compatibility, as well as verifying it works @Adrian-Stanea 

And @rkent thanks for finding the way to put it together.

PR #193, opened 2025-03-26
Rewrite theory of operation as a separate file.

rkent:
As another step in the documentation of rosdoc2, rewrite the Theory of Operation and put it in a separate file.

There are a few more PRs like this one pending in my repo, when these are mostly landed we'll clean up README.


tfoote (approved):
Thanks for the updates this definitely looks more complete and current.

PR #194, opened 2025-04-18
Link to the 'Documenting a ROS 2 package' guide

christophebedard:
I noticed that the [how-to guide](https://docs.ros.org/en/rolling/How-To-Guides/Documenting-a-ROS-2-Package.html) is not mentioned here. It has some pretty useful information, so I thought it would be good to link to it here.

Note that some of the empty/TODO sub-sections under [Documenting a Package](https://github.com/ros-infrastructure/rosdoc2#documenting-a-package) in the README here are covered by the guide. Not sure if the intention is to also document that here, but for now we can just start with this link.


sloretz (approved):

christophebedard:
I'll let you or someone else merge this, because I don't have the necessary permissions here!

PR #196, opened 2025-07-28
Extend user-provided autodoc mock imports

bjsowa:
Sometimes adding mock imports from exec dependencies is not enough as not every rosdep key translates directly to the Python package name. For example `python3-opencv` provides `cv2` package.

Currently it is impossible for the user to provide `autodoc_mock_imports` in the `conf.py` as they are overwritten by the wrapper `conf.py`. This PR modifies the wrapper `conf.py` to extend the user-provided `autodoc_mock_imports` instead of overwriting them.

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -162,7 +162,8 @@ if rosdoc2_settings.get('enable_autodoc', True):
     autodoc_mock_imports.extend(pkgs_to_mock)
 
     if len(autodoc_mock_imports) > 0:
-        print(f"[rosdoc2] autodoc mock imports: '{{'\\', \\''.join(autodoc_mock_imports)}}'")
+        joined_imports = "', '".join(autodoc_mock_imports)
+        print(f"[rosdoc2] autodoc mock imports: '{{joined_imports}}'")
 
 if rosdoc2_settings.get('enable_intersphinx', True):
     print('[rosdoc2] enabling intersphinx')


bjsowa:
I'm trying to fix the doc jobs on [rosbag2_to_video](https://github.com/fictionlab/rosbag2_to_video/tree/ros2) package (example of unstable job [here](https://build.ros2.org/user/bjsowa/my-views/view/my_packages/job/Rdoc__rosbag2_to_video__ubuntu_noble_amd64/)). With this patch, I can just add a simple [conf.py](https://github.com/fictionlab/rosbag2_to_video/blob/fix/rosdoc-warnings/doc/conf.py) which sets the mock import in `cv2` to fix the warnings.

rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -157,10 +158,10 @@ def ensure_global(name, default):
...
         except ImportError:
             pkgs_to_mock.append(exec_depend)
-    # todo(YV): If users provide autodoc_mock_imports in their conf.py
-    # it will be overwritten by those in exec_depends.
-    # Consider appending to autodoc_mock_imports instead.
-    autodoc_mock_imports = pkgs_to_mock
+    
+    autodoc_mock_imports.extend(pkgs_to_mock)
+
+    print(f"[rosdoc2] autodoc mock imports: '{{autodoc_mock_imports}}'")
```
rkent, line 164: It would be good to only print this if it is non-empty, as for the vast majority of users this is not used.

bjsowa:
> 2\. It would also be good to include this in a test. You are welcome to propose something, but if you prefer I could propose something as an addition to your commit.

I would appreciate some help with it.

tfoote (commented):
This looks good to add. Do you have some examples of packages that I can see this work on so I can validate it works?

tfoote:
I've approved the workflows. It just needs a maintainer to approve for new contributors.

tfoote (approved):
Thanks for validating this rkent. With the tests passing now I'll merge this so it gets out. And we'll look for the follow-on to get coverage.

PR #199, opened 2025-08-05
Add AUTHORS.md to standard documents

bjsowa:


rosdoc2/verbs/build/standard_documents.py
```diff
@@ -36,7 +36,15 @@
 
 def locate_standard_documents(package_xml_directory):
     """Locate standard documents."""
-    names = ['readme', 'license', 'contributing', 'changelog', 'quality_declaration', 'package']
+    names = [
```
tfoote, line 39: This makes sense to add.  For earlier readability/future maintenance could you alphabetize the list and pull it up as a constant at the top of the file.

tfoote (changes requested):

rosdoc2/verbs/build/standard_documents.py
tfoote: Thanks we haven't been the best about it in the past but as you did following PEP 8 https://peps.python.org/pep-0008/#constants with UPPER_CASE will let us converge over time.

tfoote (approved):
Thanks for the contribution and quick fixup.

PR #205, opened 2025-10-03
Show ros package dependencies.

rkent:
Generated-by: Portions of this commit may include code completion from github.copilot version 1.372.0 or later

Replaces #114 

As requested in #114, rather than detect meta packages and add dependencies only to those, this PR adds exec dependencies for all packages.

Minor issue: I'm using the python rosdistro package to determine if an exec depend is a ros package or a system dependency. Because typical documentation builds of packages rely on scanning a repo for packages, unreleased packages get added to the rosdoc2 output listing. But when those packages are scanned for exec_depends, if those packages are not released, they will not show up as ros package dependencies.

A current example of this is http://docs.ros.org/en/rolling/p/turtlebot3_applications/  This has ros package depends including turtlebot3_automatic_parking. But because that package is not released, it would not be listed in this PR as a Ros package dependency.

An earlier version that I did of this PR, rather than relying on the python package rosdistro to find valid ros packages, instead took the opposite approach of downloading rosdistro/rosdep and then excluding known system packages from the listed ros package dependencies. That would then work for unreleased ros packages, but was considerably slower than the rosdistro-based approach.

### DIFF
--- a/rosdoc2/verbs/build/builders/sphinx_builder.py
+++ b/rosdoc2/verbs/build/builders/sphinx_builder.py
@@ -22,6 +22,7 @@ import subprocess
 import sys
 
 from jinja2 import Template
+import rosdistro
 import setuptools
 
 from ..builder import Builder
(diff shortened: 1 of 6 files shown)


rosdoc2/verbs/build/builders/sphinx_builder.py
```diff
@@ -615,6 +617,28 @@ def build(self, *, doc_build_folder, output_staging_directory):
             if package_name != self.build_context.package.name
         ]
 
+        # Collect package-only exec_depends
+        exec_depends = self.build_context.package.exec_depends
+        ros_distro = os.environ.get('ROS_DISTRO')
+        if not ros_distro:
+            logger.warning('ROS_DISTRO not set, cannot check ros package dependencies')
```
tfoote: Historically we've wanted to make sure that we weren't locking in too closely onto the ROS_DISTRO. But adding it with this backwards compatibility seems like a good compromise.

PR #214, opened 2026-01-05
Remove myself from CODEOWNERS

audrow:
I'm trying to reduce my notifications in repos I'm not actively maintaining.

@tfoote, if you like I can change this PR to remove the CODEOWNERS file, so you're taken off too.


tfoote (approved):
I'm still mostly the one keeping the lights on here so I'll keep the notifications going. I'm on full overload, but this at least gets it into the github review queue for me.

PR #218, opened 2026-04-17
Fixed index rst jinja templeta

ahcorde:
Ensured a required blank line is always present between that bullet list item and the following `.. toctree::` directive.


ahcorde:
FYI @tfoote or @cottsay

tfoote (approved):
This works, if it's required there might be some logic in having the blank line after the if, not inside it, to avoid a mistake in the future. But I don't see this as a high likelihood of needing a new insertion so we don't need to fiddle.

